PR #3, opened 2018-02-23
Create directories for gnuradio blocks and flowgraphs

koncha99:


koncha99:
OK, it's more simple, I think.

PR #5, opened 2018-02-26
STELLAR-757 add GRPC server initial implementation and go project structure.

reiinakano:
Notes about this PR:

- I used the import path "github.com/infostellarinc/starcoder" since this is an open source project and it would be good to be able to `go get` it normally like any other Go library. Then we can just run `starcoder <arguments>` to run the server.

- I used the command `protoc -I starcoder-proto/ starcoder-proto/src/main/proto/starcoder.proto --go_out=plugins=grpc:starcoder-proto` to generate the `starcoder.pb.go` file in `starcoder-proto/src/main/proto/`. and I was able to import it normally.

- I realized we should default to `$HOME/.flowgraphs` (or maybe `$HOME/.starcoder/flowgraphs`?) instead of the `flowgraphs/` folder at the root repository.

Anyway, should we use Travis for CI?


starcoder.go
```diff
@@ -19,8 +19,58 @@
...
+	"log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+	"os/user"
+	"path/filepath"
+)
+
+const (
+	// TODO: Replace this with command line argument
+	port = ":50051"
```
vaelen: I would call this `bindAddress` because it can also contain an IP address to bind to.

starcoder.go
```diff
@@ -19,8 +19,58 @@
 
 package main
 
-import "fmt"
+import (
+	pb "github.com/infostellarinc/starcoder/starcoder-proto/src/main/proto"
```
anuraaga: Wait a bit, I'm going to add the proto build first (this is a weird import path for golang)

reiinakano:
Changed the import path to "github.com/infostellarinc/starcoder/protobuf" and used the command `protoc -I starcoder-proto/ starcoder-proto/src/main/proto/starcoder.proto --go_out=plugins=grpc:$GOPATH/src` to generate the proper .pb.go file.

reiinakano:
Interesting how Go suggests to check in generated Go source code in a repository.

reiinakano:
Should I do `goLock` as well?

starcoder.go
```diff
@@ -19,8 +19,58 @@
...
+		log.Fatalf("failed to get user: %v", err)
+	}
+	lis, err := net.Listen("tcp", bindAddress)
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+	s := grpc.NewServer()
+	pb.RegisterProcessManagerServer(s, &server{
+		// TODO: Replace this with command line argument
+		flowgraphDir: filepath.Join(usr.HomeDir, ".starcoder/flowgraphs"),
```
anuraaga: Can we have it default to compiled in flowgraphs (prepared by rice)?

reiinakano: You want the `.grc` files compiled into the Go binary? So default behavior is compiled in flowgraphs, but if '.starcoder/flowgraphs' exists, we use those?

anuraaga: I think the default should be compiled in unless a user specifies a flag pointing at some other flowgraphs, no need to hardcode the flowgraphs directory in our binary I think

reiinakano: Ah ok, I'll TODO this for when I add in the command line arguments library, which will probably be in the next PR.

PR #7, opened 2018-02-26
Build proto with gradle and move to api directory. 

anuraaga:
We prefer to use Gradle for building things for smarter incremental compiling and a strong ability to define build dependencies. We also don't like checking in generated code, regardless of the golang recommendations.


reiinakano:
Would `go get` still work out of the box without the generated source files?

anuraaga:
Probably not but we can think about it later - either way I'd rather just release binaries than having to build the package every time.

PR #11, opened 2018-02-27
Add dependency management.

anuraaga:


reiinakano:
Could you also add the README for updating dependencies here?

PR #12, opened 2018-02-27
STELLAR-757 Use Cobra to allow passing in of parameters

reiinakano:
Runs GRPC server with `starcoder serve -bindAddress <> -flowgraphDIr <>`

If no flowgraphDir is given, default to packed in flowgraphs.


cmd/root.go
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		// Find home directory.
+		home, err := homedir.Dir()
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		// Search config in home directory with name ".starcoder" (without extension).
+		viper.AddConfigPath(home)
```
anuraaga: How about `AddConfigPath(home/.starcoder)` and `SetConfigName(config)`?

cmd/serve.go
```diff
@@ -0,0 +1,109 @@
...
+// serveCmd represents the serve command
+var serveCmd = &cobra.Command{
+	Use:   "serve",
+	Short: "Start the Starcoder server",
+	Long: `Start the Starcoder server`,
+	Run: func(cmd *cobra.Command, args []string) {
+		log.Printf("serve called, using bind address %v", serveCmdConfig.BindAddress)
+
+		if serveCmdConfig.FlowgraphDir == "" {
+			tempDir, err := ioutil.TempDir("", "starcoder")
```
anuraaga: `defer delete the temp dir`?

build.gradle
```diff
@@ -125,6 +125,21 @@ task depGet(type: com.github.blindpirate.gogradle.Go) {
...
+task getRice(type: com.github.blindpirate.gogradle.Go) {
+    go 'get -u github.com/GeertJohan/go.rice/...'
+}
+
+task useRice(type:Exec) {
+    inputs.dir 'flowgraphs'
+    outputs.file 'cmd/rice-box.go'
+
+    workingDir project.file('cmd').toString()
+    dependsOn getRice
```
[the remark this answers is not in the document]
anuraaga: First means first (the dependency graph is the most important part of a build so we put it as the first line of a task, see the other ones)

cmd/serve.go
```diff
@@ -74,9 +78,33 @@ var serveCmd = &cobra.Command{
...
+		go func() {
+			for {
+				select {
+				case s := <-sigs:
+					switch s {
+					case syscall.SIGINT:
+						fallthrough
+					case syscall.SIGTERM:
+						fallthrough
+					case syscall.SIGQUIT:
```
anuraaga: Can you use `defer` instead?

reiinakano: Doesn't work, I don't think it gets called with SIGINT.

reiinakano: Yup, that worked.

PR #13, opened 2018-02-28
STELLAR-757 Start and end processes

reiinakano:
This PR makes Starcoder actually start and end GNURadio processes.

Some notes:
- Starcoder accepts either .grc files or already compiled .py files. If the request contains a .grc file, `grcc` is used to compile the .grc file to the corresponding .py file. All generated files are cleaned up at application exit. 
- There is still a bunch of important work left, some of which are captured in the various new TODOs, but at this point we have the bare minimum to be able to start and stop GNURadio processes through GRPC and check benchmarks on whatever hardware we want to test this on.


reiinakano:
@anuraaga should I merge this first then embed Python in a separate PR?

anuraaga:
Yeah

PR #14, opened 2018-03-02
STELLAR-818 Embed python

reiinakano:


server/starcoder.go
```diff
@@ -115,104 +118,264 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+					"argument_parser function not found. "+
+						"Cannot parse parameters. Was %v generated by GRC?", inFilePythonPath),
+			}, nil
+		} else {
+			// Top block has no parameters. Initialize with empty kwArgs
+			kwArgs = emptyTuple
+			defer safeDecRef(kwArgs)
+		}
+	} else {
+		optparseModule := python.PyImport_ImportModule("optparse")
```
anuraaga: Can't you just initialize the dict directly with the parameters? We shouldn't have to roundtrip to argparse for this.

server/starcoder.go
```diff
@@ -115,104 +118,264 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+	if callReturn == nil {
+		return &pb.StartProcessReply{
+			Status: pb.StartProcessReply_PYTHON_RUN_ERROR,
+			Error:  getExceptionString(),
+		}, nil
+	}
+
+	var uniqueId string
+	for ok := true; ok; _, ok = s.flowGraphs[uniqueId] {
+		uniqueId = util.RandString(4)
```
anuraaga: I think you can use `secureHash(flowGraphInstance.GetCPointer())` instead

server/starcoder.go
reiinakano: I want to get the types right. We could send the types through gRPC, but that would mean we'd ignore the types already set in the flowgraph itself. I'm fine either way, though.

reiinakano:
@anuraaga Do you know why the build is failing? It's fine on my local.

server/starcoder.go
anuraaga: I think your comment got deleted for some reason but yeah, model the types in the proto - since people need to know what options are in the graph to run it, the proto user will always know the type.

You can use the proto type `Value` which is one of any of the primitive types.

anuraaga:
Don't worry too much about the build right now, I'll push a new builder image today

server/starcoder.go
reiinakano: Done

api/starcoder.proto
```diff
@@ -30,6 +30,35 @@ service ProcessManager {
   rpc EndProcess (EndProcessRequest) returns (EndProcessReply) {}
 }
 
+// Complex number
+message Complex {
+  // Real value of the complex number
```
anuraaga: `Real part`

api/starcoder.proto
```diff
@@ -30,6 +30,35 @@ service ProcessManager {
...
+  double real_value = 1;
+
+  // Imaginary value of the complex number
```
anuraaga: `Imageinary part`

api/starcoder.proto
```diff
@@ -30,6 +30,35 @@ service ProcessManager {
...
+  double imaginary_value = 2;
+}
+
+// GNURadio parameter values
+message Value {
+  oneof val {
+    // Represents at string value
```
anuraaga: `a`

api/starcoder.proto
```diff
@@ -30,6 +30,35 @@ service ProcessManager {
...
+    // Represents a float
+    double float_value = 4;
+
+    // Represents a complex number
+    Complex complex_value = 5;
+  }
+}
+
 // The request message containing parameters for starting a process
 message StartProcessRequest {
```
anuraaga: Let's also go ahead and rename `Process` -> `Flowgraph`

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+	defer safeDecRef(kwArgs)
+	if kwArgs == nil {
+		return &pb.StartProcessReply{
+			Status: pb.StartProcessReply_PYTHON_RUN_ERROR,
+			Error:  getExceptionString(),
+		}, nil
+	}
+
+	for _, param := range in.GetParameters() {
+		pyKey := python.PyString_FromString(param.GetKey())
```
anuraaga: Think you can avoid this by using `PyDict_SetItemString` if golang strings can be used as `char *`

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+					Error:  err.Error(),
+				}, nil
+			}
+		default:
+			return &pb.StartProcessReply{
+				Status: pb.StartProcessReply_UNKNOWN_ERROR,
+				Error:  "Unsupported value type",
+			}, nil
+		}
+		safeDecRef(pyKey)
```
anuraaga: Put this after `FromString`

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+	}
+
+	for _, param := range in.GetParameters() {
+		pyKey := python.PyString_FromString(param.GetKey())
+		switch v := param.GetValue().GetVal().(type) {
+		case *pb.Value_StringValue:
+			err = python.PyDict_SetItem(
+				kwArgs,
+				pyKey,
+				python.PyString_FromString(v.StringValue))
```
anuraaga: Need to decrement the reference on this value throughout, `PyDict_SetItem` adds a reference. Have the switch statement just create `Pyobject` for the value and then after the switch, call `PyDict_SetItem`

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+
+	flowGraphInstance := flowgraphClass.Call(emptyTuple, kwArgs)
+	if flowGraphInstance == nil {
+		return &pb.StartProcessReply{
+			Status: pb.StartProcessReply_PYTHON_RUN_ERROR,
+			Error:  getExceptionString(),
+		}, nil
+	}
+
+	callReturn := flowGraphInstance.CallMethod("start")
```
anuraaga: Decret the return (it's `None`, so not a big deal but still)

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
-			},
+			Status: pb.EndProcessReply_PYTHON_RUN_ERROR,
+			Error:  getExceptionString(),
 		}, nil
 	}
-
-	err := cmd.Process.Signal(os.Interrupt)
-	if err != nil {
+	// TODO: Is it possible "stop" won't work? Should we timeout "wait"?
+	callReturn = flowGraph.CallMethod("wait")
```
anuraaga: I think you'll eventually be listening for a message from your sink block, or something. Calling `wait` probably doesn't matter much.

server/starcoder.go
```diff
@@ -115,104 +119,271 @@ func (s *Starcoder) StartProcess(ctx context.Context, in *pb.StartProcessRequest
...
+		safeDecRef(flowGraph)
 	}
 	for _, tempDir := range s.temporaryDirs {
 		os.RemoveAll(tempDir)
 	}
 	return nil
 }
+
+func safeDecRef(obj *python.PyObject) {
+	if obj != nil {
```
anuraaga: I think you basically always check for nil, so no need to have this `safeDecRef` method

cmd/serve.go
```diff
@@ -104,6 +105,7 @@ var serveCmd = &cobra.Command{
 						log.Println("Caught signal", sig)
 						starcoder.Close()
 						s.GracefulStop()
+						python.Finalize()
```
anuraaga: How about `defer` this similar to removing the tmp directory. You might still want the flowgraph to run during the graceful shutdown.

server/starcoder.go
reiinakano: I saw this but I don't think the go-python library has `PyDict_SetItemString` that takes the `char *`. I might be misunderstanding the documentation, not sure how to access it if it is present. https://godoc.org/github.com/sbinet/go-python#PyDict_SetItem

anuraaga: Not a big deal if it's not there

server/starcoder.go
reiinakano: I get the warning "possible resource leak, `defer` called in a for loop" from Idea, though. Should I just ignore?

anuraaga: I guess since `defer` runs on the function, not block, and it would only cause the last value to defer. I guess in this case, we want
```go
go fun() {
  defer
}()
```

server/starcoder.go
reiinakano: It's not ideal, but there are certain blocks for which we may need to wait for the flowgraph to actually close so we know when to collect some generated files. I would eventually like to rewrite them so it will just notify the client via a message, but I think we could keep this for now.

server/starcoder.go
reiinakano: The reason I made this is so I can keep the `defer` immediately after creation. Otherwise I'd have to do 
```go
obj := createPythonObj
if obj == nil {
  return ...
}
defer obj.DecRef()
```
vs

```go
obj := createPythonObj
defer safedec(obj)
if obj == nil {
  return ...
}
```

No strong opinions on this though.

server/starcoder.go
reiinakano: Not sure I follow. This would create a goroutine that would exit almost immediately, decref-ing `pykey`, probably before I've used it, right?

reiinakano: Did some searching, the suggested way is to just close at the end of the loop without deferring https://stackoverflow.com/questions/45617758/defer-in-the-loop-what-will-be-better

reiinakano: Ah maybe not a goroutine but just a regular anon function?

api/starcoder.proto
```diff
@@ -24,14 +24,43 @@ option go_package = "/api";
 // The GNURadio process manager service definition.
 service ProcessManager {
   // Create a process
-  rpc StartProcess (StartProcessRequest) returns (StartProcessReply) {}
+  rpc StartProcess (StartFlowgraphRequest) returns (StartFlowgraphReply) {}
```
anuraaga: `StartFlowgraph` for method name too, and we use the convention of `Response` instead of `Reply`

server/starcoder.go
anuraaga: Right thinking about it at that point is fine

server/starcoder.go
anuraaga: Yeah anon function sounds good

reiinakano:
I think I've addressed everything.

PR #15, opened 2018-03-05
Decrement references

reiinakano:
Looks like I missed DECREFing some references last PR.


server/starcoder.go
```diff
@@ -124,6 +124,13 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
 	// Append module directory to sys.path
 	log.Printf("Appending %v to sys.path", filepath.Dir(inFilePythonPath))
 	sysPath := python.PySys_GetObject("path")
+	if sysPath == nil {
+		return &pb.StartFlowgraphResponse{
+			Status: pb.StartFlowgraphResponse_PYTHON_RUN_ERROR,
+			Error:  getExceptionString(),
+		}, nil
+	}
+	defer sysPath.DecRef()
```
anuraaga: This is a borrowed reference so shouldn't need to decref
https://docs.python.org/2/c-api/sys.html

PR #16, opened 2018-03-05
STELLAR-758 Add AR2300 source block

koncha99:
Added the source block for supporting input from AR2300.
(AR2300 is the digital receiver developed by AOR, Ltd.  http://www.aorusa.com/receivers/ar2300.html)


gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,168 @@
...
+    throw std::runtime_error("ar2300_receiver::initialize");
+  }
+
+  init_flag = true;
+}
+
+/*
+ * Stop receiving
+ */
+void ar2300_receiver::stop() {
```
gavinreaney: Naming: it would be better to have the name be symmetrical with the one above: e.g.

start, stop
create, destroy
initialize, finalize

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,168 @@
...
+  if (context != NULL) {
+    libusb_exit(context);
+    context = NULL;
+  }
+
+  init_flag = false;
+}
+
+/*
+ * Read IQ data from device
```
gavinreaney: Say something about the meaning of the return code.

gr-starcoder/lib/ar2300_driver.h
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) 2018 InfoStellar, Inc. All Rights Reserved.
```
gavinreaney: Do we need to have the open source licence blurb here? Same for the other files too.

gr-starcoder/lib/ar2300_receiver.h
```diff
@@ -0,0 +1,46 @@
...
+#include <libusb-1.0/libusb.h>
+#include "ar2300_driver.h"
+}
+
+class ar2300_receiver {
+ public:
+  ar2300_receiver();
+  ~ar2300_receiver();
+
+  void initialize();
```
gavinreaney: Is the code still a WIP? We should have more comments here I guess.

It might be worth mentioning if there are threading restrictions.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+      receiver->initialize();
+    }
+
+    /*
+     * Our virtual destructor.
+     */
+    ar2300_source_impl::~ar2300_source_impl()
+    {
+      receiver->stop();
+      if (buf != NULL) {
```
gavinreaney: Pretty sure you don't need the NULL check (delete should be safe with NULL).

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+      int outSize = encode_ar2300(buf, ret, out);
+
+      // Tell runtime system how many output items we produced.
+      return outSize;
+    }
+
+    int
+    ar2300_source_impl::encode_ar2300(char* in,
```
gavinreaney: const char*

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+
+      if (!seenIValue) {
+        return 0;
+      }
+
+      int out_index = 0;
+      for (int i = offset; i < inSize; ++i) {
+        sample[sample_index++] = in[i];
```
gavinreaney: I was wondering why sample is a member of the class rather than a local. Is it to allow for the case where a sample crosses a read boundary?

Are you certain that 'offset' is correct in the case where a sample crosses a boundary? I couldn't quite convince myself about that.

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,272 @@
...
+extern volatile int event_thread_run;
+
+/* write isochronous data to file
+ *
+ * Internal function to write out the received isochronous data
+ * out to disk.
+ *
+ * Returns the number of bytes unwritten or -1
+ */
+int iq_packet_write(AR2300_HANDLE *ar2300, unsigned char *buffer, int length) {
```
gavinreaney: const unsigned char *

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,272 @@
...
+  int left;
+  ssize_t written;
+
+  left = length;
+  while (left > 0) {
+    written = write(ar2300->outfd, buffer, left);
+    if (written <= 0) {
```
gavinreaney: I think you set the FD to non-blocking, so a return code of -1 is expected in cases where the pipe is full. Is that really something we want to treat as an error condition?

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,272 @@
...
+ * out to disk.
+ *
+ * Returns the number of bytes unwritten or -1
+ */
+int iq_packet_write(AR2300_HANDLE *ar2300, unsigned char *buffer, int length) {
+  int left;
+  ssize_t written;
+
+  left = length;
+  while (left > 0) {
```
gavinreaney: Do we ever loop? It seems that unless everything is written in one go, then we return at line 46.

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,272 @@
...
+          if (bytes_left > 0) {
+            /* some error occurred during write */
+            ar2300->err_func(transfer, AR2300_ERR_INCOMPLETE_WRITE);
+          }
+          if (bytes_left < 0) {
+            ar2300->err_func(transfer, AR2300_ERR_DATA_WRITE);
+          }
+        } else {
+          /* some error occurred for this packet */
+          LOGGER(LOG_DEBUG, "packet %d status %d %d %d\n", idx, hdr->status,
```
gavinreaney: Do we want to keep logging this special case? It seems strange that the other error conditions don't use LOGGER.

(This error case does happen every few seconds in my testing).

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,272 @@
...
+      break;
+    default: {
+      /* some serious error occurred during transfer */
+      ar2300->err_func(transfer, AR2300_ERR_ISO_STATUS);
+    }
+  }
+
+  /* reuse this transfer */
+  if (ar2300->iso_status == AR2300_ISO_READY) {
+    result = libusb_submit_transfer(transfer);
```
gavinreaney: It's an issue with the existing code, but if this fails, I think the driver is basically broken until the higher level software restarts it.

gr-starcoder/lib/ar2300_driver.h
```diff
@@ -0,0 +1,246 @@
...
+
+/** @file */
+#ifndef AR2300_DRIVER_H
```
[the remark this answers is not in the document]
koncha99: I think so. I move #endif to the end of file.

gr-starcoder/lib/driver.c
koncha99: I will address in following commit.

gr-starcoder/lib/ar2300_driver.h
```diff
@@ -0,0 +1,247 @@
...
+/** could not write complete data to output */
+#define AR2300_ERR_INCOMPLETE_WRITE (8)
+
+/** error during data write */
+#define AR2300_ERR_DATA_WRITE (9)
+
+/** transfer error for this iso packet */
+#define AR2300_ERR_ISO_PACKET (10)
+
+#endif
```
anuraaga: Newline at end of file throughout

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+      return outSize;
+    }
+
+    int
+    ar2300_source_impl::encode_ar2300(char* in,
+                              const int inSize,
+                              gr_complex* out)
+    {
+      int offset = 0;
+      if (!seenIValue) {
```
anuraaga: Is it ok to keep this state between multiple reads? What happens if we miss some data due to could load, etc.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+
+      return out_index;
+    }
+
+    gr_complex
+    ar2300_source_impl::parse_sample() const
+    {
+      float real = ((sample[0] << 24) | (sample[1] << 16 & 16646144) |
```
anuraaga: It might be better to use hex for these magic numbers

gr-starcoder/lib/ar2300_source_impl.h
```diff
@@ -0,0 +1,59 @@
...
+
+      int encode_ar2300(char* in, int size, gr_complex* out);
+      gr_complex parse_sample() const;
+
+     public:
+      ar2300_source_impl();
+      ~ar2300_source_impl();
+
+      // Where all the action really happens
+      int work(int noutput_items,
```
anuraaga: Let's use modern c++ where possible, I think this is `override`

gr-starcoder/python/__init__.py
```diff
@@ -22,13 +22,12 @@
 This is the GNU Radio STARCODER module. Place your Python package
 description here (python/__init__.py).
 '''
-
 # import swig generated symbols into the starcoder namespace
 try:
 	# this might fail if the module is python-only
 	from starcoder_swig import *
```
anuraaga: Just wondering, do we need to generate the py file even for use from golang? Is it not possible to use a pure-C++ block in a Python flowgraph?

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+
+    int
+    ar2300_source_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int ret = receiver->read(buf, buf_size, timeout_ms);
+      if (ret == 0) {
+        return noutput_items;
```
reiinakano: If there were no produced items, I think `work` must return 0. https://wiki.gnuradio.org/index.php/BlocksCodingGuide#The_work_function

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+      int outSize = encode_ar2300(buf, ret, out);
```
reiinakano: `n_output_items` contains the available space in this block's output buffer for the current `work` call. I'm not familiar with C++, but is it possible that you're writing more than `n_output_items` samples to `out`? Can `outSize > n_output_items`?

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+      }
+    }
+
+    int
+    ar2300_source_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int ret = receiver->read(buf, buf_size, timeout_ms);
```
reiinakano: Since this is a blocking call inside `work`, does it follow the guidelines here https://wiki.gnuradio.org/index.php/BlocksCodingGuide#Blocking_calls? 

Do you think it's possible to convert `work` to non-blocking? Like perhaps check quickly if the AR2300 pipe contains new samples then put those in `out` (up to `n_output_items` only), and if it doesn't contain samples, just return. I'm not familiar with the driver so maybe it's not possible, just something to keep in mind.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,128 @@
...
+
+    int
+    ar2300_source_impl::encode_ar2300(char* in,
+                              const int inSize,
+                              gr_complex* out)
+    {
+      int offset = 0;
+      if (!seenIValue) {
+        for (int i = 1; i < inSize; i += 2) {
+          if (in[i] & 0x1) {
```
reiinakano: I think it's better to do the checks here all the time, not just once at the very beginning, in case of corruption. Also, might be better to do all the checks for the 4 words, not just the 1st one, again in case of corruption.

Also, is it guaranteed that the `in` bytes won't be off by 1? e.g.
Expected:
`in = xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 ...`

Offby1 (Is this possible?):
`in = xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 ...`

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: You are right. Number of samples do not exceed ``` noutput_items```. I fixed it.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Done.

gr-starcoder/lib/ar2300_driver.h
koncha99: Done.

gr-starcoder/lib/ar2300_receiver.cc
koncha99: Changed to "start", "stop".

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: You are right. I changed it to local variable.

gr-starcoder/lib/ar2300_source_impl.h
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,122 @@
...
+      receiver->stop();
+    }
+
+    int
+    ar2300_source_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int buf_size = noutput_items;
```
reiinakano: Since there are 8 bytes per complex sample, should this be `int buf_size = noutput_items*8`?

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: This is called with timeout and this return immediately if operation is timeout. So, I'm still not sure it's best practice but align to guideline.

reiinakano: Sounds good

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Fixed.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: I added validation for each sample. I can't think of any ideas which cause latter case except something software glitches around reading logic. 
I assume there is no broken samples between device and pipe except for entire sample dropping.

gr-starcoder/lib/driver.c
koncha99: Can I separate this bug as other issue? We have already found another bug.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+        if (in[i] & 0x1) {
+          offset = i - 1;
+          break;
+        }
+        return 0;
+      }
+
+      int out_index = 0;
+      for (int i = offset; i < inSize; ++i) {
+        sample[sample_index++] = in[i];
```
reiinakano: So far LGTM. One question: Is it guaranteed that there won't be an incomplete sample at the end of `in`? e.g.

`in = .... xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx`

If this happens often then you'd be dropping samples, correct?

gr-starcoder/python/__init__.py
reiinakano: The Python file is just used so the flowgraph can be put together in Python. During runtime, there are no performance hits as the block and scheduler runs in pure C++. https://wiki.gnuradio.org/index.php/FAQ#Why_use_Python.3F_It.E2.80.99s_so_slow.21

gr-starcoder/lib/driver.c
gavinreaney: Yes

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    int
+    ar2300_source_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int buf_size = noutput_items * 8;
+      char* buf = new char[buf_size];
```
anuraaga: Don't think this is freed. Anyways it doesn't look like this is called from multiple threads so can reuse a static buffer.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+      return outSize;
+    }
+
+    int
+    ar2300_source_impl::encode_ar2300(const char* in,
+                              const int inSize,
+                              gr_complex* out)
+    {
+      char  sample[8];
+      int   sample_index = 0;
```
anuraaga: Initialize this below closer to where it is used

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+          if(!validate_sample(sample)) {
+            GR_LOG_WARN(d_logger, boost::format("Sample may be corrupted"));
+            num_of_consecutive_warns++;
+            if(num_of_consecutive_warns > CONSECUTIVE_WARNING_LIMIT) {
+              GR_LOG_WARN(d_logger, boost::format("Exceeded the limit of consecutive warnings"));
+              throw std::runtime_error("ar2300_source_impl::encode_ar2300()");
+            }
+            continue;
+          }
+          gr_complex value = parse_sample(sample);
```
anuraaga: Don't need this local

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+    ar2300_source_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int buf_size = noutput_items * 8;
+      char* buf = new char[buf_size];
+
+      int ret = receiver->read(buf, buf_size, timeout_ms);
+      if (ret == 0) {
```
anuraaga: `ret < 8`

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,141 @@
...
+    /*
+     * Our virtual destructor.
+     */
+    ar2300_source_impl::~ar2300_source_impl()
+    {
+      receiver->stop();
+    }
+
+    int
+    ar2300_source_impl::work(int noutput_items,
```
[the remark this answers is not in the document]
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: I added buffer deletion at the end of function. But, is it efficient?

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Currently, yes. Incomplete sample will be dropped.

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,185 @@
...
+  if (stat < 0) {
+    stop();
+    fprintf(stderr, "ar2300_receiver::read: invalid status was returned. stat=%d\n", stat);
+    throw std::runtime_error("ar2300_receiver::read");
+  } else if (stat == 0) {
+    // timeout
+    return 0;
+  }
+
+  int ret = ::read(read_pipe[PIPE_READ], buf, size);
```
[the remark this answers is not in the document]
koncha99: I agree with that. I will file it as a issue.

gr-starcoder/lib/ar2300_source_impl.cc
reiinakano: Okay, maybe we can leave this as a TODO for now in case too many samples are dropped over the duration of a pass. Would be good to store any incomplete sample and put it together in the next call to `work`.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
+      int buf_size = n_output_items * 8;
+      char* buf = new char[buf_size];
+
+      int ret = receiver->read(buf, buf_size, timeout_ms);
+      if (ret < 8) {
+        return 0;
```
gavinreaney: buf is leaked if you exit here.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+    {
+      receiver->stop();
+    }
+
+    int
+    ar2300_source_impl::work(int n_output_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      gr_complex *out = (gr_complex *) output_items[0];
```
gavinreaney: Just checking my understanding:
gr_vector_void_star is a vector<void *> with one entry,. And that one entry is an array of gr_complex which can hold n_output_items?

It's kind of ugly that they didn't allow us to fill a vector<gr_complex> as output and avoid all the risk of overwriting the buffer.

As an aside: have you ever used memory debugging tools like valgrind? Might be good to have it enabled in debug or test builds.

reiinakano: I know almost nothing about C++ but the base class `block` is designed to take different types of inputs/outputs, not just `gr_complex`. Maybe that's why? I don't know if there's a better way they could've designed this though :)

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -0,0 +1,142 @@
...
+      int buf_size = n_output_items * 8;
+      char* buf = new char[buf_size];
```
anuraaga: It's ok not to use a static buf, but in that case at least change it to a stack variable for easy management. Or if that doesn't work since buf_size is too large, than use `unique_ptr`

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: n_output_items is handled by gnuradio runtime. Return value of work call will be used by next block.
(Does this answer your question?)

And thank you for your advise! I'm trying to apply valgrind or sanitizer for debug build.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -125,7 +125,7 @@ namespace gr {
       float imag = ((in[4] << 24) | (in[5] << 16 & 0xFE0000) |
                                   (in[6] << 9 & 0x1FE00) | (in[7] << 1 & 0x1FC)) >>
                                  2;
-      return gr_complex(real, imag);
+      return gr_complex(real * powf(10, -7), imag * powf(10, -7));
```
reiinakano: Will this be optimized into a constant by the compiler? If not, then it might be more efficient to just define a constant `AR2300_SCALE_FACTOR` or something.

koncha99: Done.

gr-starcoder/lib/ar2300_source_impl.cc
koncha99: I tested simple flow graph(qa_starcoder::run_ar2300_source_block) with valgrind. It didn't showed any lost.

gavinreaney: Great :)

gr-starcoder/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
anuraaga: Can you change the top-level gitignore to build instead of `/build/` instead?

koncha99: Done.

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,185 @@
...
+    read_pipe[PIPE_READ] = -1;
+    read_pipe[PIPE_WRITE] = -1;
+  }
+
+  if (context != NULL) {
+    libusb_exit(context);
+    context = NULL;
+  }
+
+  init_flag = false;
```
gavinreaney: Could maybe call this flag started instead of init_flag. Up to you.

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/*
+ * Read IQ data from device
+ * @return: 0  - Timeout
+ *          >0 - Number of bytes read
```
gavinreaney: Is it possible to have no timeout but read zero bytes?

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -0,0 +1,185 @@
...
+  struct timeval tv;
+  tv.tv_sec = (int)(timeout / 1000.0);
+  tv.tv_usec = (timeout % 1000) * 1000;
+
+  // Initialize file descriptor
+  fd_set rfds;
+  FD_ZERO(&rfds);
+  FD_SET(read_pipe[PIPE_READ], &rfds);
+
+  int ret = ::select(FD_SETSIZE, &rfds, (fd_set*)NULL, (fd_set*)NULL, &tv);
```
gavinreaney: May as well just return ::select(...

gr-starcoder/lib/driver.c
```diff
@@ -0,0 +1,284 @@
...
+  if (written <= 0) {
+    if (errno != 0) {
+      return -1;
+    } else {
+      return left;
+    }
+  }
+
+  left -= written;
+  buffer += written;
```
gavinreaney: I guess this line isn't needed.

koncha99: Done.

gr-starcoder/lib/ar2300_receiver.cc
koncha99: Done.

gr-starcoder/lib/ar2300_receiver.cc
koncha99: Yes, you are right. I added more description.

gr-starcoder/lib/ar2300_receiver.cc
koncha99: It seems to good. I changed it.

koncha99:
I added some modification to this opensourced version. So, I will sync it to internal version. 
After that, Fix will be applied internal version first and then synced to starcoder until completing switching to starcoder.

PR #17, opened 2018-03-05
properly release the GIL in between Python calls

reiinakano:


server/starcoder.go
```diff
@@ -121,6 +123,11 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
 	}
 	// TODO: Have some way to monitor the running process
 
+	python.PyEval_RestoreThread(s.threadState)
```
anuraaga: I think you also want to lock the os thread, with multiple syscalls happening at the same time, there is no guarantee that the consecutive Python calls run on the same thread, which might mess it up.

reiinakano:
I followed the pattern suggested by https://github.com/golang/go/wiki/LockOSThread

server/python_runtime.go
```diff
@@ -0,0 +1,78 @@
...
+
+func NewPythonRuntime() *PythonRuntime {
+	p := &PythonRuntime{
+		funcChannel:  make(chan func()),
+		closeChannel: make(chan bool),
+	}
+
+	go func() {
+		// Lock all Python calls to a single OS thread
+		runtime.LockOSThread()
```
anuraaga: I think python should be ok running on any OS thread as long as it's the same thread throughout a block that keeps the GIL, unlike OpenGL etc. Can you move these to surround `f()` instead of the processing loop? Otherwise, no other goroutine will be schedulable onto this thread ever, and that's probably inefficient.

reiinakano: Ahh I see, in that case, it might be simpler to do:

```go
StartFlowgraph() {
lock os thread
python.RestoreThread
//python stuff
SaveThread
unlock os thread
}
```

wdyt?

anuraaga: Yeah actually that's what I was expecting :)

PR #18, opened 2018-03-06
Claim GIL and lock OS thread during Python calls

reiinakano:


server/starcoder.go
```diff
@@ -121,6 +124,13 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
 	}
 	// TODO: Have some way to monitor the running process
 
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
anuraaga: Put the `UnlockOsThread` in the same defer, since we want to make sure it is called after `SaveThread`. You can define a helper method that does the two operations.

server/starcoder.go
```diff
@@ -347,6 +364,8 @@ func (s *Starcoder) EndFlowgraph(ctx context.Context, in *pb.EndFlowgraphRequest
 }
 
 func (s *Starcoder) Close() error {
+	runtime.LockOSThread()
+	python.PyEval_RestoreThread(s.threadState)
```
anuraaga: Make sure to cleanup here too (not as necessary but we always should)

server/starcoder.go
reiinakano: I believe `defer`'s are last in first out.

server/starcoder.go
reiinakano: I did `runtime.UnlockOSThread` but calling `PyEval_SaveThread` seems to break the `python.Finalize` call

anuraaga: Ah, `Finalize` will also need to be on the GIL. It's a bit weird though that we have a thread dependency between this method and the finalize call - what do you think about initializing python in `newStarCoderServer` instead then, so the python thread is 100% tied to this object?

reiinakano: I moved the Python initialization to the Starcoder level, which probably makes more sense.

PR #20, opened 2018-03-12
STELLAR-759 gRPC message sink

reiinakano:
Contains GRPC message sink for sending PMT messages to any gRPC endpoint. Since we already have the AR2300 source block, this is all we need for now.


gr-grpc/.gitignore
```diff
@@ -0,0 +1,4 @@
+lib/*.pb.cc
```
anuraaga: Just put these in the top level .gitignore

gr-grpc/api/grpcapi.proto
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+syntax = "proto3";
+
+package grpcapi;
+
+// The PMT Message sink
+service MessageSink {
+  // Send a serialized PMT message
+  rpc SendMessage (SendMessageRequest) returns (SendMessageResponse) {}
```
anuraaga: semicolon instead of empty method body

gr-grpc/api/grpcapi.proto
```diff
@@ -0,0 +1,47 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+syntax = "proto3";
+
+package grpcapi;
+
+// The PMT Message sink
```
anuraaga: Define PMT

reiinakano:
This OOT is patterned after https://github.com/gnuradio/gnuradio/tree/master/gr-zeromq, but instead of ZMQ, it uses gRPC. The role of this block is to get info out of the flowgraph (which is done using sink blocks), which right now is only possible through files (file sink), or ZMQ (zmq sink).

anuraaga:
Yeah, instead can you get the data into golang? We want the golang grpc server to be the interface to the world, which is a different design than writing straight to a message queue. A client can connect with a streaming gRPC API to the starcoder and get bytes back on the stream, the client shouldn't have to have it's own gRPC server to listen for packets

reiinakano:
I see. Yeah, that is a different design. I'll look into how I can do that

PR #21, opened 2018-03-13
Enqueue msg sink

reiinakano:
#### gr-starcoder changes:
Adds a block called `enqueue_msg_sink`. Provides method to grab a queue which can then be polled for new entries, so you can do anything with them (sent through gRPC, in our case).

Written in Python for now for simplicity, but performance-wise, this isn't really a bottleneck for us since it is out of the path of GNURadio streaming blocks (it only handles async messages, not samples).

#### Starcoder changes:
Searches the flowgraph for `enqueue_msg_sink` blocks. User can then call `StreamPmt` for a particular block to get a stream of the messages from that block.

### DIFF
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -28,6 +28,9 @@ service ProcessManager {
 
   // End a process
   rpc EndFlowgraph (EndFlowgraphRequest) returns (EndFlowgraphResponse) {}
+
+  // Stream from enqueue message sink block
+  rpc StreamPmt (StreamPmtRequest) returns (stream StreamPmtResponse) {}
 }
 
 // Complex number
@@ -141,4 +144,19 @@ message EndFlowgraphResponse {
 
   // Error message
   string error = 2;
+}
+
+// The message containing a serialized PMT (GNURadio Polymorphic Message Type)
+message StreamPmtRequest {
+  // Flowgraph ID
+  string processId = 1;
+
+  // Block ID for specifying which enqueue_msg_sink block to stream
+  string blockId = 2;
+}
+
+// The message containing a serialized PMT (GNURadio Polymorphic Message Type)
+message StreamPmtResponse {
+  // Serialized arbitrary PMT message
+  bytes bytes = 1;
 }
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -22,7 +22,7 @@ package starcoder;
 option go_package = "/api";
 
 // The GNURadio process manager service definition.
-service StarCoder {
+service Starcoder {
   // Runs a flowgraph and streams back
   rpc RunFlowgraph (stream RunFlowgraphRequest) returns (stream RunFlowgraphResponse) {}
 }
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -23,14 +23,36 @@ option go_package = "/api";
 
 // The GNURadio process manager service definition.
 service ProcessManager {
-  // Create a process
-  rpc StartFlowgraph (StartFlowgraphRequest) returns (StartFlowgraphResponse) {}
+  // Runs a flowgraph and streams back
+  rpc RunFlowgraph (stream RunFlowgraphRequest) returns (stream RunFlowgraphResponse) {}
+}
 
-  // End a process
-  rpc EndFlowgraph (EndFlowgraphRequest) returns (EndFlowgraphResponse) {}
+message RunFlowgraphRequest {
+  // Filename of GNURadio script to run
+  string filename = 1;
 
-  // Stream from enqueue message sink block
-  rpc StreamPmt (StreamPmtRequest) returns (stream StreamPmtResponse) {}
+  // Key value pair representing a parameter to be passed to the script
+  message Parameter {
+    // Name of parameter
+    string key = 1;
+
+    // Value to assign the parameter
+    Value value = 2;
+  }
+
+  // Arbitrary command line parameters to be passed to the GNURadio script
+  repeated Parameter parameters = 2;
+
+  // Blocks to stream back in the response
+  repeated string block_id = 3;
+}
+
+message RunFlowgraphResponse {
+  // Block ID to which this response belongs
+  string block_id = 1;
+
+  // Serialized arbitrary PMT (GNURadio Polymorphic Message Type)
+  bytes payload = 2;
 }
 
 // Complex number
@@ -60,103 +82,4 @@ message Value {
     // Represents a complex number
     Complex complex_value = 5;
   }
-}
-
-// The request message containing parameters for starting a flowgraph
-message StartFlowgraphRequest {
-  // Filename of GNURadio script to run
-  string filename = 1;
-
-  // Key value pair representing a parameter to be passed to the script
-  message Parameter {
-    // Name of parameter
-    string key = 1;
-
-    // Value to assign the parameter
-    Value value = 2;
-  }
-
-  // Arbitrary command line parameters to be passed to the GNURadio script
-  repeated Parameter parameters = 2;
-}
-
-// The message response to a StartFlowgraphRequest
-message StartFlowgraphResponse {
-  // Unique Process ID of the flowgraph that was started
-  string processId = 1;
-
-  // Response status
-  enum Status {
-    // Successfully started flowgraph.
-    SUCCESS = 0;
-
-    // File access problem
-    FILE_ACCESS_ERROR = 1;
-
-    // Unsupported file type
-    UNSUPPORTED_FILE_TYPE = 2;
-
-    // .grc compile error
-    GRC_COMPILE_ERROR = 3;
-
-    // Python file run error
-    PYTHON_RUN_ERROR = 4;
-
-    // Unknown error starting flowgraph.
-    UNKNOWN_ERROR = 255;
-  }
-
-  // Response status
-  Status status = 2;
-
-  // Error message
-  string error = 3;
-}
-
-// The request message containing parameters for ending a flowgraph
-message EndFlowgraphRequest {
-  // Process ID of process to end
-  string processId = 1;
-}
-
-// The message response to an EndFlowgraphRequest
-message EndFlowgraphResponse {
-  // Response status
-  enum Status {
-    // Successfully ended flowgraph.
-    SUCCESS = 0;
-
-    // Invalid Process Id
-    INVALID_PROCESS_ID = 1;
-
-    // Flowgraph has exited
-    PROCESS_EXITED = 2;
-
-    // Python file run error
-    PYTHON_RUN_ERROR = 3;
-
-    // Unknown error ending flowgraph.
-    UNKNOWN_ERROR = 255;
-  }
-
-  // Response status
-  Status status = 1;
-
-  // Error message
-  string error = 2;
-}
-
-// The message containing a serialized PMT (GNURadio Polymorphic Message Type)
-message StreamPmtRequest {
-  // Flowgraph ID
-  string processId = 1;
-
-  // Block ID for specifying which enqueue_msg_sink block to stream
-  string blockId = 2;
-}
-
-// The message containing a serialized PMT (GNURadio Polymorphic Message Type)
-message StreamPmtResponse {
-  // Serialized arbitrary PMT message
-  bytes bytes = 1;
 }
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -21,12 +21,6 @@ package starcoder;
 
 option go_package = "/api";
 
-// The GNURadio process manager service definition.
-service Starcoder {
-  // Runs a flowgraph and streams back
-  rpc RunFlowgraph (stream RunFlowgraphRequest) returns (stream RunFlowgraphResponse) {}
-}
-
 message RunFlowgraphRequest {
   // Filename of GNURadio script to run
   string filename = 1;
@@ -82,4 +76,10 @@ message Value {
     // Represents a complex number
     Complex complex_value = 5;
   }
+}
+
+// The GNURadio process manager service definition.
+service Starcoder {
+  // Runs a flowgraph and streams back
+  rpc RunFlowgraph (stream RunFlowgraphRequest) returns (stream RunFlowgraphResponse) {}
 }
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -22,7 +22,7 @@ package starcoder;
 option go_package = "/api";
 
 // The GNURadio process manager service definition.
-service ProcessManager {
+service StarCoder {
   // Runs a flowgraph and streams back
   rpc RunFlowgraph (stream RunFlowgraphRequest) returns (stream RunFlowgraphResponse) {}
 }
--- a/gr-starcoder/python/enqueue_msg_sink.py
+++ b/gr-starcoder/python/enqueue_msg_sink.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2018 <+YOU OR YOUR COMPANY+>.
+# 
+# This is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+# 
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+import numpy
+from gnuradio import gr
+import pmt
+from collections import deque
+
+class enqueue_msg_sink(gr.basic_block):
+    """
+    docstring for block enqueue_msg_sink
+    """
+    def __init__(self):
+        gr.basic_block.__init__(self,
+            name="enqueue_msg_sink",
+            in_sig=[],
+            out_sig=[])
+        self.q = deque()
+        self.message_port_register_in(pmt.intern("in"))
+        self.set_msg_handler(pmt.intern("in"), self.msg_handler)
+
+    def get_q(self):
+        return self.q
+
+    def msg_handler(self, pmt):
+        self.q.append(pmt)
+        print("Received pmt", pmt)
--- a/gr-starcoder/python/enqueue_msg_sink.py
+++ b/gr-starcoder/python/enqueue_msg_sink.py
@@ -32,12 +32,16 @@ class enqueue_msg_sink(gr.basic_block):
             name="enqueue_msg_sink",
             in_sig=[],
             out_sig=[])
-        self.q = deque()
         self.message_port_register_in(pmt.intern("in"))
         self.set_msg_handler(pmt.intern("in"), self.msg_handler)
+        self.observers = []
 
-    def get_q(self):
-        return self.q
+    def observe(self):
+        q = deque()
+        self.observers.append(q)
+        return q
 
     def msg_handler(self, message):
-        self.q.append(bytearray(pmt.serialize_str(message)))
+        serialized_bytes = bytearray(pmt.serialize_str(message))
+        for q in self.observers:
+            q.append(serialized_bytes)
--- a/gr-starcoder/python/enqueue_msg_sink.py
+++ b/gr-starcoder/python/enqueue_msg_sink.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # 
-# Copyright 2018 <+YOU OR YOUR COMPANY+>.
+# Copyright 2018 Infostellar.
 # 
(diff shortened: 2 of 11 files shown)


gr-starcoder/python/enqueue_msg_sink.py
```diff
@@ -0,0 +1,47 @@
...
+import pmt
+from collections import deque
+
+class enqueue_msg_sink(gr.basic_block):
+    """
+    docstring for block enqueue_msg_sink
+    """
+    def __init__(self):
+        gr.basic_block.__init__(self,
+            name="enqueue_msg_sink",
```
anuraaga: single-quotes for strings

server/starcoder.go
```diff
@@ -315,6 +376,46 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+			if flowGraphDict == nil {
+				return &pb.StartFlowgraphResponse{
+					Status: pb.StartFlowgraphResponse_PYTHON_RUN_ERROR,
+					Error:  getExceptionString(),
+				}, nil
+			}
+			defer flowGraphDict.DecRef()
+
+			iter := 0
+			key := python.PyString_FromString("")
```
anuraaga: I think `var key: PyObject`, otherwise you leak these two strings you created

gr-starcoder/python/enqueue_msg_sink.py
```diff
@@ -0,0 +1,47 @@
...
+            in_sig=[],
+            out_sig=[])
+        self.message_port_register_in(pmt.intern("in"))
+        self.set_msg_handler(pmt.intern("in"), self.msg_handler)
+        self.observers = []
+
+    def observe(self):
+        q = deque()
```
anuraaga: You don't need a `deque`, just use a list `[]`

server/starcoder.go
```diff
@@ -323,6 +424,104 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+				runtime.LockOSThread()
+				python.PyEval_RestoreThread(s.threadState)
+				defer func() {
+					s.threadState = python.PyEval_SaveThread()
+					runtime.UnlockOSThread()
+				}()
+				pyLength := pmtQueue.CallMethod("__len__")
+				length := python.PyInt_AsLong(pyLength)
+
+				emptyTuple := python.PyTuple_New(0)
```
anuraaga: Don't think this is used

server/starcoder.go
```diff
@@ -323,6 +424,104 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+				if emptyTuple == nil {
+					return nil, errors.New(getExceptionString())
+				}
+				defer emptyTuple.DecRef()
+
+				var bytes [][]byte
+
+				for i := 0; i < length; i++ {
```
anuraaga:
```golang
length := PyList_GET_SIZE(queue)
for i := 0; i < length; i++ {
  pmtBytes = PyByteArray_AsBytes(PyList_GET_ITEM(queue, i)
}
if PyList_SetSlice(queue, 0, length, NULL) == -1 {
  PythonRunError
}
```
Not how if we use a list, we can avoid reference counting by using the ref-stealing version of methods

server/starcoder.go
reiinakano: Errors out when I just do this, so I just `decref`'ed them.

api/starcoder.proto
```diff
@@ -23,11 +23,36 @@ option go_package = "/api";
 
 // The GNURadio process manager service definition.
 service ProcessManager {
```
anuraaga: `StarCoder`

server/starcoder.go
```diff
@@ -122,276 +110,313 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
-	}
-	defer topBlock.DecRef()
-
-	// Verify top_block subclass
-	isSubclass := flowgraphClass.IsSubclass(topBlock)
-	if isSubclass == 0 {
-		return &pb.StartFlowgraphResponse{
-			Status: pb.StartFlowgraphResponse_PYTHON_RUN_ERROR,
-			Error: fmt.Sprintf(
+		emptyTuple := python.PyTuple_New(0)
```
anuraaga: Initialize this closer to where you use it

server/starcoder.go
```diff
@@ -122,276 +110,313 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+		}
 	}()
 
-	flowGraph := s.flowGraphs[in.GetProcessId()]
+	// Streaming loop here
+	err = func() error {
+		for {
+			select {
+			case <-closeChannel:
+				return nil
```
anuraaga: I think you want to make sure all the bytes have been transferred. I guess you call `stop` here and wait for the graph to finish, make sure all bytes are sent, and then close the response.

server/starcoder.go
```diff
@@ -122,276 +110,313 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+						var bytes [][]byte
+
+						emptyList := python.PyList_New(0)
+						defer emptyList.DecRef()
+
+						for i := 0; i < length; i++ {
+							// TODO: Convert PMT to a gRPC native data structure.
+							// Use built-in PMT serialization for now.
+							pmtBytes := python.PyByteArray_AsBytes(python.PyList_GetItem(pmtQueue, i))
+							err := python.PyList_SetSlice(pmtQueue, 0, length, emptyList)
```
anuraaga: So `nil` doesn't work where the python API expects NULL? :(

server/starcoder.go
```diff
@@ -122,276 +110,313 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+							if err != nil {
+								return nil, err
+							}
+							log.Println(k, pmtBytes)
```
anuraaga: Do you still need this? Looks like lots of spam

server/starcoder.go
reiinakano: It doesn't :/

reiinakano:
Addressed comments

server/starcoder.go
```diff
@@ -50,30 +49,31 @@ func NewStarcoderServer(flowgraphDir string) *Starcoder {
 	return &Starcoder{
 		flowgraphDir:  flowgraphDir,
 		temporaryDirs: make([]string, 0),
-		flowGraphs:    make(map[string]*python.PyObject),
 		threadState:   threadState,
 	}
 }
 
-func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphRequest) (*pb.StartFlowgraphResponse, error) {
+func (s *Starcoder) RunFlowgraph(stream pb.StarCoder_RunFlowgraphServer) error {
```
anuraaga: Let's make the case of C consistent between code and gRPC stub. I have no preference, maybe lowercase c was better but go ahead and pick one.

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
 
 func (s *Starcoder) Close() error {
+
 	runtime.LockOSThread()
 	python.PyEval_RestoreThread(s.threadState)
 	defer func() {
 		python.Finalize()
 		runtime.UnlockOSThread()
 	}()
-	for _, flowGraph := range s.flowGraphs {
```
anuraaga: I think you still need to cleanup if any streams are in progress, stopping flowgraphs and terminating gRPC streams

server/starcoder.go
anuraaga: Can you split this method up? It's way too huge.

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+	python.PyEval_RestoreThread(s.threadState)
+	defer func() {
+		s.threadState = python.PyEval_SaveThread()
+		runtime.UnlockOSThread()
+	}()
+	length := python.PyList_Size(pmtQueue)
+
+	var bytes [][]byte
+
+	emptyList := python.PyList_New(0)
```
anuraaga: What about `import "C"` and C.NULL? Or just `0`?

anuraaga: Might also need to use unsafe.Pointer and/or uintptr

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
 			}
-			return nil
-		}()
-		if response != nil {
-			return response, nil
 		}
+		return flowGraphInstance, msgSinkBlocks, nil
+	}()
+	if err != nil {
+		return err
```
anuraaga: I'm not too sure how this err is handled by you should fail the gRPC stream with some information from this err object

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+		}
+	}()
 
-	s.flowGraphs[uniqueId] = flowGraphInstance
+	// Streaming loop here
+	err = func() error {
+		for {
+			select {
+			case <-closeChannel:
+				err := s.stopFlowGraph(flowGraphInstance)
```
anuraaga: This seems to block the goroutine which is bad if doing multiple graphs at the same time. I think we wanted to get rid of `wait` later which would be needed to solve this, so if you want to fix it later leave a TODO

gr-starcoder/python/enqueue_msg_sink.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2018 <+YOU OR YOUR COMPANY+>.
```
vaelen: Infostellar

gr-starcoder/python/enqueue_msg_sink.py
```diff
@@ -0,0 +1,46 @@
...
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+from gnuradio import gr
+import pmt
+
+class enqueue_msg_sink(gr.basic_block):
+    """
+    docstring for block enqueue_msg_sink
```
vaelen: You should go ahead and add a doc string.

gr-starcoder/python/qa_enqueue_msg_sink.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2018 <+YOU OR YOUR COMPANY+>.
```
vaelen: Infostellar

sampleclient/main.go
```diff
@@ -0,0 +1,71 @@
...
+func main() {
+	conn, err := grpc.Dial("127.0.0.1:8990", grpc.WithInsecure())
+	if err != nil {
+		log.Fatalf("fail to dial: %v", err)
+	}
+	defer conn.Close()
+	client := pb.NewStarCoderClient(conn)
+
+	stream, err := client.RunFlowgraph(context.Background())
+	waitc := make(chan struct{})
```
vaelen: This works, but have you considered a WaitGroup?

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
 	}
-	// TODO: Have some way to monitor the running process
 
-	runtime.LockOSThread()
-	python.PyEval_RestoreThread(s.threadState)
-	defer func() {
-		s.threadState = python.PyEval_SaveThread()
-		runtime.UnlockOSThread()
-	}()
+	flowGraphInstance, msgSinkBlocks, err := func() (*python.PyObject, map[string]*python.PyObject, error) {
```
vaelen: I suggest moving this function out from inside the outer function.  This will make it easier to test and it will lower the complexity of the outer function.

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+		runtime.LockOSThread()
```
vaelen: It would be good to add a comment here.  I assume this is because two python interpreters can't run in the same OS thread?

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+		python.PyEval_RestoreThread(s.threadState)
+		defer func() {
+			s.threadState = python.PyEval_SaveThread()
+			runtime.UnlockOSThread()
```
vaelen: You should probably defer `UnlockOSThread` before running `PyEval_RestoreThread`, I think.  Defers are put into a LIFO queue, so the order should be correct.

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+			return nil, nil, errors.New(getExceptionString())
+		}
+
+		kwArgs := python.PyDict_New()
+		if kwArgs == nil {
+			return nil, nil, errors.New(getExceptionString())
+		}
+		defer kwArgs.DecRef()
+
+		for _, param := range in.GetParameters() {
```
vaelen: This would make a good method by itself.

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+			if err != nil {
+				closeChannel <- true
+				log.Printf("%v", err)
+				return
+			}
+		}
+	}()
 
-	s.flowGraphs[uniqueId] = flowGraphInstance
+	// Streaming loop here
```
vaelen: I suggest breaking this out into an actual function.  Besides making the outer function smaller, it makes debugging easier because it gives a real name to the function.

server/starcoder.go
reiinakano: I think returning an `error` from the `RunFlowgraph` function is enough to be handled by gRPC. The error will be seen on the client

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
 }
 
-func (s *Starcoder) EndFlowgraph(ctx context.Context, in *pb.EndFlowgraphRequest) (*pb.EndFlowgraphResponse, error) {
-	if _, ok := s.flowGraphs[in.GetProcessId()]; !ok {
-		return &pb.EndFlowgraphResponse{
-			Status: pb.EndFlowgraphResponse_INVALID_PROCESS_ID,
-			Error:  fmt.Sprintf("Invalid process ID %v", in.GetProcessId()),
-		}, nil
+func (s *Starcoder) getBytesFromQueue(pmtQueue *python.PyObject) ([][]byte, error) {
+	runtime.LockOSThread()
```
vaelen: To be clear, this is a second OS thread that we're locking besides the one locked in the previous function, so we are now using 2 cores that can't be used for any other goroutines, correct?

server/starcoder.go
```diff
@@ -122,276 +110,333 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
+		if err != nil {
+			return nil, err
+		}
+		bytes = append(bytes, pmtBytes)
 	}
+	return bytes, nil
+}
 
+func (s *Starcoder) stopFlowGraph(flowGraphInstance *python.PyObject) error {
 	runtime.LockOSThread()
```
vaelen: Third core.

sampleclient/main.go
reiinakano: I didn't pay too much attention to this. This is from the gRPC docs.

server/starcoder.go
reiinakano: https://github.com/infostellarinc/starcoder/pull/18/files/bbc794203808c25335f0b31367442fe2403251d1#diff-e50d0cc1b7352fd33b195c315750ba02

Which one is better?

anuraaga: I like having the two in a single function that's deferred still, it's easier to reason about. Would make sense to have a helper method though.

server/starcoder.go
reiinakano: It is from this comment: https://github.com/infostellarinc/starcoder/pull/17/files/7b48aa899c0e7d758d5fc8e17d67478a4d715c7b#diff-e50d0cc1b7352fd33b195c315750ba02
I've ran it without `LockOSThread` before without problems, but I guess it's better to be safe? Don't know enough about Python internals to know for sure.

server/starcoder.go
vaelen: My concern is if PyEval_RestoreThread panics the unlock won't get called.

server/starcoder.go
reiinakano: No, the function that calls this is not holding the GIL, otherwise it would lock up.

anuraaga: We only lock when making subsequent python calls and then unlock when done, so it should never lock more than one core. There's also no requirement that the same OS thread runs the python calls as long as the same thread runs all calls between a `RestoreThread` and `SaveThread`, so it seemed simpler to me than trying to manage python calls via a goroutine with channels. Do you see a problem with the current pattern?

sampleclient/main.go
vaelen: They both solve the same problem, so I have no strong opinion.  The WaitGroup seems more obvious to me, especially if someone doesn't know Go well.

server/starcoder.go
reiinakano: Ditto. These don't run at the same time.

server/starcoder.go
vaelen: Here are the language details, just for reference: https://blog.golang.org/defer-panic-and-recover

sampleclient/main.go
reiinakano: Btw, do you think I should just remove this? Might be a bit messy to commit into the actual repo, but I'm not sure

server/starcoder.go
vaelen: Nope, that sounds good to me, that just wasn't obvious to me from looking at the code the first time.

vaelen: You could make a function like this to make the code cleaner:
```
func WithPythonInterpreter(f func()) {
  // setup interpreter
  f()
  // cleanup interpreter
}
```

server/starcoder.go
anuraaga: If `RestoreThread` panics, the GIL will be an unknown state and this program will almost assuredly be dead - I don't think anyone would be able to embed a python interpreter if that was practically possible ;) The most important point for me is having a single fun with the two calls, which should be extracted, deferring within the defer seems ok too, but is probably overkill due to the above point. Actually, now I wonder if we use `defer` more than we should in general as it seems to have a reasonable overhead (why isn't it implemented by the compiler?).

https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32

sampleclient/main.go
vaelen: Nah, it's fine.

server/starcoder.go
vaelen: We should probably add a top level defer then that calls recover and kills the whole process if it catches something.

server/starcoder.go
reiinakano: Neither seems to work. I tried accessing the actual C function `C.PyList_SetSlice` instead of the one given by `go-python` but I can't convert `pmtQueue` into its `*C.Pyobject` form since it's an internal variable.

reiinakano:
I think I've addressed the comments

server/starcoder.go
```diff
@@ -268,130 +489,93 @@ func (s *Starcoder) StartFlowgraph(ctx context.Context, in *pb.StartFlowgraphReq
...
-	if waitCallReturn == nil {
-		return &pb.EndFlowgraphResponse{
-			Status: pb.EndFlowgraphResponse_PYTHON_RUN_ERROR,
-			Error:  getExceptionString(),
-		}, nil
+	for i := 0; i < length; i++ {
+		// TODO: Convert PMT to a gRPC native data structure.
+		// Use built-in PMT serialization for now.
+		pmtBytes := python.PyByteArray_AsBytes(python.PyList_GetItem(pmtQueue, i))
+		err := python.PyList_SetSlice(pmtQueue, 0, length, emptyList)
```
anuraaga: I forget if I already commented this but can't find it. It's too bad `nil` doesn't work, but I think it should still be possible to pass `NULL` somehow. `C.NULL`, `uintptr(0)`, something like that.

reiinakano: I've already filed the relevant issue here: https://github.com/sbinet/go-python/issues/61

I skimmed the source code a bit and the problem is it expects a `*python.PyObject`, which is a wrapper over the C object. It would probably work if I passed in the `C.NULL` object at that level, but it's an internal variable so I can't change it.

PR #24, opened 2018-03-30
Further cleanup of CW+FSK flowgraphs

reiinakano:
Also:
Removed gr-grpc since we don't need this anymore
Added test.grc flowgraph for basic testing w/ no need for dummy IQ data


reiinakano:
Sorry, which configs? I you're talking about the config file starcoder uses when running then its already in yanl

anuraaga:
From looking at gnuradio's repos I found that it supports yaml instead of xml for flowgraphs now adays from what I could tell

reiinakano:
Oh interesting. I think they're still in the process of migrating though. https://wiki.gnuradio.org/index.php/OldDevCalls and https://github.com/gnuradio/gnuradio/pull/1481 seem to suggest this will be a 3.8 feature. Since the current stable release is 3.7, we should probably wait until the stable release grc accepts yaml instead of XML. As a side note, 3.8 will also support py3 so we can just sneak this in along with that bigger change.

PR #25, opened 2018-04-09
implementation of complex_to_msg_c block

reiinakano:
This GNURadio block monitors an incoming stream of complex floats and sends them as a PMT message when `num_items` items are received. This is so we can still send partial.bin files to GCP

One thing of note that is left as a TODO is that this implementation may drop up to `num_items - 1` items at the very end when the flowgraph is closed. For our use case, this means up to 1MB of data (since we will set it to send 1MB messages). 

From calculations, 1MB of data is equivalent to ~1.165 seconds of data at the very end for a 112500Hz sample rate. `(1024^2 bytes) / (8 bytes/sample) / (112500samples/sec) = ~1.165secs`

At higher sampling rates (>UHF), the amount of time dropped will be even lower. Those last few milliseconds aren't likely to hold any useful information since the satellite will be gone by then. Note that this only affects the partial IQ data this block sends.


flowgraphs/test.grc
```diff
@@ -587,6 +618,49 @@
...
+      <key>maxoutbuf</key>
+      <value>0</value>
+    </param>
+    <param>
+      <key>minoutbuf</key>
+      <value>0</value>
+    </param>
+    <param>
+      <key>num_items</key>
+      <value>131072</value>
```
gavinreaney: What's the significance of this value? (I mean how was it chosen and would we ever want to change it?).

flowgraphs/test.grc
```diff
@@ -738,4 +812,16 @@
...
+  <connection>
+    <source_block_id>blocks_throttle_0</source_block_id>
+    <sink_block_id>starcoder_complex_to_msg_c_0</sink_block_id>
+    <source_key>0</source_key>
+    <sink_key>0</sink_key>
+  </connection>
+  <connection>
+    <source_block_id>starcoder_complex_to_msg_c_0</source_block_id>
+    <sink_block_id>partial_iq_sink</sink_block_id>
+    <source_key>out</source_key>
```
gavinreaney: Reading this confused me at first...(I'd expect something that's a source to be called 'in').

However, I think I get it now: the source is the thing that comes 'out' of starcoder_complex_to_msg_c_0. Is that understanding correct?

gr-starcoder/grc/starcoder_complex_to_msg_c.xml
```diff
@@ -0,0 +1,33 @@
...
+  <category>[starcoder]</category>
+  <import>import starcoder</import>
+  <make>starcoder.complex_to_msg_c($num_items)</make>
+  <!-- Make one 'param' node for every Parameter you want settable from the GUI.
+       Sub-nodes:
+       * name
+       * key (makes the value accessible as $keyname, e.g. in the make node)
+       * type -->
+  <param>
+    <name>Number of items</name>
```
gavinreaney: Is this descriptive enough?

Would it be better as something like 'Number of items in each published message' (or something like that - you can think of a better wording).

gr-starcoder/include/starcoder/complex_to_msg_c.h
```diff
@@ -0,0 +1,56 @@
+/* -*- c++ -*- */
+/* 
+ * Copyright 2018 <+YOU OR YOUR COMPANY+>.
```
gavinreaney: I assume we don't need to put infostellar here?

gr-starcoder/include/starcoder/complex_to_msg_c.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace gr {
+  namespace starcoder {
+
+    /*!
+     * This block takes in a stream of complex floats and packs them together
+     * into a PMT blob containing `num_items` items.
+     *
+     */
+    class STARCODER_API complex_to_msg_c : virtual public gr::sync_block
```
gavinreaney: Using virtual inheritence is pretty unusual ... but I see the gnu radio example does it too (I'm OK to keep it but I think it isn't required right?).

gr-starcoder/lib/complex_to_msg_c_impl.cc
```diff
@@ -0,0 +1,73 @@
...
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      const gr_complex *in = (const gr_complex *) input_items[0];
+
+      for (int i = 0; i < noutput_items/num_items; i++) {
+        message_port_pub(pmt::mp("out"), pmt::mp(in+(i*num_items), sizeof(gr_complex)*num_items));
+      }
+
+      return noutput_items;
```
gavinreaney: Should we actually return 0 here? I'm a bit unclear on what happens to the 'output_items' since we didn't actually write anything there.

gr-starcoder/lib/complex_to_msg_c_impl.cc
```diff
@@ -0,0 +1,73 @@
...
+
+    int
+    complex_to_msg_c_impl::work(int noutput_items,
+        gr_vector_const_void_star &input_items,
+        gr_vector_void_star &output_items)
+    {
+      const gr_complex *in = (const gr_complex *) input_items[0];
+
+      for (int i = 0; i < noutput_items/num_items; i++) {
+        message_port_pub(pmt::mp("out"), pmt::mp(in+(i*num_items), sizeof(gr_complex)*num_items));
```
gavinreaney: Do these message hang around in memory forever until something consumes them? (Not saying anything is wrong  but just making sure I understand how it works).

flowgraphs/test.grc
reiinakano: 131072*8bytes = 1048576bytes = 1MB

flowgraphs/test.grc
reiinakano: Yep. It makes sense when you see the flow graph in GNURadio companion. These xml files aren't meant to be read. I don't even read them. :)

gr-starcoder/include/starcoder/complex_to_msg_c.h
reiinakano: Oops :)

gr-starcoder/include/starcoder/complex_to_msg_c.h
reiinakano: This was autogenerated by the tool GNURadio provided to help write these blocks. Honestly the only thing I modify here is the `work` function.

gr-starcoder/lib/complex_to_msg_c_impl.cc
anuraaga: Is it possible to output a proto instead of pmt here? Now that we're wiring up starpass and starcoder we should think about the interface in more detail - it's better if clients don't have to worry about gnuradio's details, so returning a proto from starcoder is good. We can return a proto here or convert in the starcoder - but the latter would require extra overhead even though we're just serializing - deserializing - reserializing

gr-starcoder/lib/complex_to_msg_c_impl.cc
reiinakano: A sink block like this is simply a "sync" (synchronous) block with 0 output ports. We still need to tell the GNURadio scheduler that we output `noutput_items` so it knows we consumed the input. 

It's a bit confusing since there are 0 output ports, but it makes sense when you consider there can be X number of ports, and a synchronous block means that for each input sample, there will be X corresponding outputs.

gr-starcoder/lib/complex_to_msg_c_impl.cc
reiinakano: Yeah, probably. The block after this consumes these messages. The GNURadio scheduler handles all this message passing so I assume they take care of memory leaks as well.

reiinakano: @anuraaga these values are ultimately returned by starcoder as a proto to the client. These pmt messages go to the Enqueue message sink blocks we built earlier and into the protobuf we send back to the client.

reiinakano: I've actually already successfully uploaded these partial bin files to the cloud without changing anything in starpass.

anuraaga: Are the pmts in the same format as what we are creating in starpass already? If so then sounds good

reiinakano: Yep

vaelen:
The data loss at the end of the pass seems fine to me.  We can always start and/or stop the pass sooner and/or later as needed to compensate.

flowgraphs/test.grc
vaelen: I like that this is configurable.

gr-starcoder/python/qa_complex_to_msg_c.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2018 <+YOU OR YOUR COMPANY+>.
```
vaelen: Miss one?

gr-starcoder/python/qa_complex_to_msg_c.py
```diff
@@ -0,0 +1,41 @@
...
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+from gnuradio import gr, gr_unittest
+from gnuradio import blocks
+import starcoder_swig as starcoder
+
+class qa_complex_to_msg_c (gr_unittest.TestCase):
```
anuraaga: As this is a fairly simple block (no external dependencies like AR2300) how about writing tests now?

gr-starcoder/lib/complex_to_msg_c_impl.cc
reiinakano: Although I think I get your point about not using the PMT format. Right now it works since we always just treat the data as binary data and use the default PMT format for an arbitrary blob (which just appends 8 bytes at the start we can remove when we upload to cloud). Maybe eventually, we'll want to be smarter than just treating everything as raw binary data and sending it to the starcoder client, but for our current use cases, this seems sufficient. In the future, we may want to have a discussion about generating a proper protobuf from a PMT, especially when we work with third party blocks generating complex PMTs we can't just serialize trivially (none so far, though!).

gr-starcoder/python/qa_complex_to_msg_c.py
reiinakano: Done

gr-starcoder/python/qa_complex_to_msg_c.py
```diff
@@ -0,0 +1,73 @@
...
+        expected = np.array([0, 0, 0, 0, 0, 0, 128, 63, 0, 0, 128, 63, 0, 0, 0, 64], dtype='uint8')
+
+        src = blocks.vector_source_c(src_data)
+        ctm = starcoder.complex_to_msg_c(2)
+        snk = blocks.message_debug()
+        self.tb.connect(src, ctm)
+        self.tb.msg_connect((ctm, 'out'), (snk, 'store'))
+        self.tb.run()
+
+        assert snk.num_messages() == 1
```
anuraaga: I assume `gr_unittest.TestCase` extends `unittest.TestCase` so you should have `self.assert*` methods available

reiinakano: Done

PR #27, opened 2018-04-17
STELLAR-1021 Add our own Waterfall SInk

reiinakano:
This change adds our own Waterfall Sink heavily based on the sink from Satnogs. All you really need to review is "waterfall_plotter_impl.cc". Some differences:

1. The `waterfall_plotter` sub-block is now written in C++. This lets us avoid continuous `np.concatenate` calls that may be inefficient. This is now replaced by storing a linked list of pointers to the buffers passed in to `work`. When the flowgraph is stopped, we combine the saved buffers into one and pass it to a `matplotlib` function for generating the .png image.
2. More importantly, we can choose whether to save the .png in the filesystem as before and/or save the .png bytes in memory so the flow graph caller can extract them after the flow graph is closed. This will let Starcoder get the .png bytes from the block through a public function (not yet implemented) and pass it back through gRPC.

I kept the license headers for Satnogs in the `waterfall_heatmap` and `waterfall_sink` hierarchical block since we mostly just copied those.


gr-starcoder/python/plot_waterfall.py
```diff
@@ -0,0 +1,43 @@
...
+    plt.figure(dpi=400)
+    fig = plt.imshow(arr, cmap=cm.nipy_spectral,
+                    interpolation='none', vmin=-120, vmax=-70, alpha=0.8,
+                    aspect='auto')
+
+    fig.axes.get_xaxis().set_visible(False)
+    fig.axes.get_yaxis().set_visible(False)
+    plt.axis('off')
+
+    buf = io.BytesIO()
```
anuraaga: You should pass in `initial_bytes` (I guess `fig` must provide some way of knowing it, otherwise the calculation is of course 8 * 4 * width * height for RGBA)

gr-starcoder/python/plot_waterfall.py
```diff
@@ -0,0 +1,43 @@
...
+    plt.savefig(buf, bbox_inches='tight', pad_inches=0)
```
anuraaga: Can we explicitly set the format? Not sure what format this saves as

gr-starcoder/python/plot_waterfall.py
```diff
@@ -0,0 +1,43 @@
...
+    if filename is not None:
```
anuraaga: Let's do this before the above `savefig` since no need to split the two pieces of logic

gr-starcoder/python/qa_waterfall_heatmap.py
```diff
@@ -0,0 +1,41 @@
...
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+from gnuradio import gr, gr_unittest
+from gnuradio import blocks
+import starcoder_swig as starcoder
+
+class qa_waterfall_heatmap (gr_unittest.TestCase):
```
anuraaga: Let's write tests

gr-starcoder/python/qa_waterfall_plotter.py
```diff
@@ -0,0 +1,47 @@
...
+        self.tb = None
+
+    def test_001_t (self):
+        src_data = tuple(range(140, 180)*20)
+        src = blocks.vector_source_b(src_data)
+        s2v = blocks.stream_to_vector(gr.sizeof_char, self.fft_size)
+        op = starcoder.waterfall_plotter(self.fft_size, "")
+        self.tb.connect(src, s2v, op)
+        self.tb.run()
+        # check data
```
anuraaga: Let's do this

gr-starcoder/lib/waterfall_plotter_impl.cc
```diff
@@ -0,0 +1,177 @@
...
+
+      PyGILState_STATE gstate;
+      gstate = PyGILState_Ensure();
+      d_init_numpy_array();
+
+      npy_intp dims[2]{
+        static_cast<long int>(d_total_size / (d_fft_size * sizeof(int8_t))),
+        static_cast<long int>(d_fft_size)};
+      const int ND = 2;
+      std::cout << "dims: " << dims[0] << " " << dims[1] << std::endl;
```
anuraaga: Is it normal to have printfs in gnuradio block src?

gr-starcoder/lib/waterfall_plotter_impl.h
```diff
@@ -0,0 +1,60 @@
...
+  namespace starcoder {
+    struct item {
+      size_t size;
+      char *arr;
+    };
+
+    class waterfall_plotter_impl : public waterfall_plotter
+    {
+     private:
+      std::list<item> d_list_of_arrays;
```
anuraaga: What does `d_` stand for? I've only ever seen the `m_` prefix

gr-starcoder/lib/waterfall_plotter_impl.cc
```diff
@@ -0,0 +1,177 @@
...
+
+    /*
+     * The private constructor
+     */
+    waterfall_plotter_impl::waterfall_plotter_impl(size_t fft_size, char* filename)
+      : gr::sync_block("waterfall_plotter",
+              gr::io_signature::make(1, 1, fft_size * sizeof(int8_t)),
+              gr::io_signature::make(0, 0, 0)),
+        d_total_size(0),
+        d_blob(new char[0]),
```
anuraaga: Not sure if this leaks an empty array or not (which would be leaking the array header with no content)

gr-starcoder/lib/waterfall_plotter_impl.cc
```diff
@@ -0,0 +1,177 @@
...
+
+      Py_ssize_t image_size;
+      image_size = PyString_Size(result);
+      std::cout << image_size << " = Retrieved image size" << std::endl;
+      char *image_buffer;
+      image_buffer = PyString_AsString(result);
+      if (image_buffer == NULL)
+        goto error;
+
+      d_blob = new char[image_size];
```
anuraaga: Are you intending to hook `d_blob` into a message output later? Until we use it, I think we can leave it out as it's confusing

gr-starcoder/lib/waterfall_heatmap_impl.h
```diff
@@ -0,0 +1,86 @@
...
+
+
+      int
+      general_work (int noutput_items,
+                    gr_vector_int &ninput_items,
+                    gr_vector_const_void_star &input_items,
+                    gr_vector_void_star &output_items);
+
+      size_t
+      compute_decimation(int8_t *out, const gr_complex *in, size_t n_fft);
```
anuraaga: Think these helpers can be `private`

gr-starcoder/lib/waterfall_heatmap_impl.cc
```diff
@@ -0,0 +1,298 @@
+/* -*- c++ -*- */
```
anuraaga: Add our license as well to copied files

`
/*
Infostellar license
*/
/*
sagnogs license
*/
`

gr-starcoder/lib/waterfall_plotter_impl.h
reiinakano: Dunno. But it's here under naming conventions https://wiki.gnuradio.org/index.php/Coding_guide_impl

gr-starcoder/lib/waterfall_plotter_impl.cc
reiinakano: Actually, this is a workaround for not being able to use message passing. As soon as the flow graph is closed, all the message passing blocks are also closed. This is the same reason I can't return the <1MB of data from the previous PR.

But for other stuff, like demodulated packets, etc., that are sent *while* the flowgraph is running, I already use enqueue_msg_sink which works fine.

anuraaga: As discussed offline, if this lets all the other blocks run so we get all pending data, it seems like an ok, but sort of hacky, solution. It'd be good to make sure that and make sure there isn't a cleaner way though by going through the APIs once

reiinakano: I opened an issue here: https://github.com/gnuradio/gnuradio/issues/1735 in case the maintainers can help point us in the right direction.

I examined the API again and didn't see anything that could help us. 

One solution I'm considering is this: Currently, we get data from `enqueue_msg_sink` blocks by calling their `observe` method and grabbing a queue, into which they send their messages (which is just raw binary data). Perhaps instead of detecting `enqueue_msg_sink` blocks, we could just detect if a block has an `observe` method (maybe rename to `starcoder_observe` to prevent conflict). This waterfall block can then just drop the final .png into its queue. Technically, it's not really different from what we're planning to do, but at least we won't need to change the Starcoder code. 

The only problem really is that for this special case, we're sending a message while closing the flowgraph. At this point, the message can't be picked up by another block anymore. For any other messages we send *during* runtime (decoded packets, partial.bins), we can just wire it up to an `enqueue_msg_sink` block as designed, and it'll automatically send it out without needing to do anything special.

anuraaga: I was hoping `enqueue_msg_sink` is just a stopgap since polling a queue is reasonably inefficient - we should be able to directly handle messages without polling. I was thinking that at some point, we'd use message-passing to allow directly handling them instead of polling the queue which is why I'm interested in finding a way to make it work.

reiinakano: I've thought about this too but I get stuck when I think about how I can pass a Go callback to a block  since it has to go through Python too. I like Go but I wonder if we overcomplicated things by using Go instead of just Python.

anuraaga: Much of the world is migrating from Python -> Go since the latter provides similar expressiveness with much better performance, so I see starcoder as a way to bring this to gnuradio too. While it's not trivial to interface golang to gnuradio, I think it's worth doing and think it shouldn't be super-hard - all the python just wraps c++ classes, meaning you must be able to access the C++ API directly from golang as well (maybe with SWIG_ConvertPtr or https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/lib/hier_block2.cc#L66) - I don't see any reason it wouldn't work with some swig files and glue code.

But it's ok to think about it while moving along with stopgaps like enqueue_msg_sink, and maybe even this `stop` method, still we should see how we can make the binding more golang-native

gr-starcoder/python/plot_waterfall.py
anuraaga: There's  also a way to convert matplotlib into a numpy array more directly (I guess just two copies, while this has an unknown number of copies due to growing arraay)

gr-starcoder/lib/waterfall_plotter_impl.cc
reiinakano: Some blocks do, some blocks only print when some setting is verbose, and some have printfs commented out. Will just remove them from this, I'm just using them for debug anyway.

reiinakano:
I've added the tests and addressed comments. I also modified the plotter to add grid labels, etc.

Sample image from one of our passes:
wut

gr-starcoder/python/qa_waterfall_plotter.py
```diff
@@ -0,0 +1,54 @@
...
+from gnuradio import blocks
+import starcoder_swig as starcoder
+import os
+
+class qa_waterfall_plotter (gr_unittest.TestCase):
+
+    def setUp (self):
+        self.tb = gr.top_block ()
+        self.fft_size = 40
+        self.filename = "waterfall.png"
```
anuraaga: Can you set this to a `tempfile.mktemp`?

reiinakano: Done

PR #28, opened 2018-04-20
Add support for building gr-starcoder with the builder's gnuradio pre…

anuraaga:
…fix.

- Removed setting the pipe size since for some reason I couldn't get it to build, but we're currently working on removing it so think it's ok
- Not too sure why I had to add FindLog4cpp - did we run the gnuradio codegen without support for it? Was having problems linking against a gnuradio that includes log4cpp without it

### DIFF
--- a/gr-starcoder/CMakeLists.txt
+++ b/gr-starcoder/CMakeLists.txt
@@ -133,6 +133,7 @@ endif(APPLE)
 # Find gnuradio build dependencies
 ########################################################################
 find_package(CppUnit)
+find_package(Log4cpp)
 find_package(Doxygen)
 
 # Search for GNU Radio and its components and versions. Add any
@@ -161,6 +162,7 @@ endif(DOXYGEN_FOUND)
 # Set up Python
 ########################################################################
 find_package(PythonLibs 2 REQUIRED)
+find_package(NumPy REQUIRED)
 
 ########################################################################
 # Setup the include and linker paths
@@ -174,6 +176,7 @@ include_directories(
     ${CPPUNIT_INCLUDE_DIRS}
     ${GNURADIO_ALL_INCLUDE_DIRS}
     ${PYTHON_INCLUDE_DIRS}
+    ${NumPy_INCLUDE_DIRS}
 )
 
 link_directories(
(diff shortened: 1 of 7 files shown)


gr-starcoder/CMakeLists.txt
```diff
@@ -161,6 +162,7 @@ endif(DOXYGEN_FOUND)
 # Set up Python
 ########################################################################
 find_package(PythonLibs 2 REQUIRED)
+find_package(NumPy REQUIRED)
```
reiinakano: Do you know why we need this in the build but I didn't need it to build locally? I just assumed it worked because my Python installation had Numpy installed. But not too sure...

gr-starcoder/lib/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ if(NOT starcoder_sources)
 endif(NOT starcoder_sources)
 
 add_library(gnuradio-starcoder SHARED ${starcoder_sources})
-target_link_libraries(gnuradio-starcoder ${Boost_LIBRARIES} ${GNURADIO_ALL_LIBRARIES} usb-1.0)
+target_link_libraries(gnuradio-starcoder ${Boost_LIBRARIES} ${GNURADIO_ALL_LIBRARIES} ${LOG4CPP_LIBRARY} usb-1.0)
```
reiinakano: I don't think we use this in gr-starcoder. Do we also need this to build?

gr-starcoder/CMakeLists.txt
anuraaga: I think you installed it into your OS python and you're just using that. CMake generally falls back to OS packages fine but won't find anything in a different environment, like a pyenv or condaenv. Ubuntu 18.04 is coming out soon and I'd recommend you reinstall at that point and avoid installing system packages since it makes it very hard to make reproducible builds - most python users use environments of some sort anyways.

gr-starcoder/lib/CMakeLists.txt
anuraaga: When starcoder links against gnuradio, which links against it, we have to mention it here. Technically starcoder uses it, since it links against gnuradio's log commands, which use log4cpp behind the scenes.

There seems to be a closed, but unresolved, issue related to it, I guess `GNURADIO_ALL_LIBRARIES` should contain it but it doesn't due to an upstream bug.

https://github.com/gnuradio/gnuradio/issues/1045

PR #30, opened 2018-04-23
STELLAR-1044 Replace OS pipe in AR2300 driver with blocking queue.

reiinakano:
Replaces the OS pipe with a special queue that blocks during `pop` until data is available.

### DIFF
--- a/gr-starcoder/lib/ar2300_driver.h
+++ b/gr-starcoder/lib/ar2300_driver.h
@@ -23,7 +23,7 @@
 
 #define AR2300_USE_PTHREAD 1 /**< include threading code for event handler */
 #define AR2300_USE_SYSLOG 1  /**< output debug messages to syslog */
-
+#include "starcoder_queue.h"
 /*
  * constants related to receiver spec
  */
@@ -117,6 +117,8 @@ typedef struct ar2300_handle {
 
   int outfd; /**< I/Q output file descriptor */
 
+  starcoder_queue* q; // Boost ring buffer for storing IQ data
+
   transfer_error_callback_func err_func; /**< error callback */
 
   /** prepare multiple iso transfers so that
@@ -195,6 +197,11 @@ int ar2300_stop_transfer(AR2300_HANDLE *ar2300);
  * @param fd the file to write I/Q data to
  */
 void ar2300_set_fd(AR2300_HANDLE *ar2300, int fd);
+
+/**
+ * set the ring buffer
+ */
+void ar2300_set_q(AR2300_HANDLE *ar2300, starcoder_queue *q);
 
 /**
  * start an event handling thread
--- a/gr-starcoder/lib/ar2300_driver.h
+++ b/gr-starcoder/lib/ar2300_driver.h
@@ -115,8 +115,6 @@ typedef struct ar2300_handle {
 
   ar2300_bulk_status bulk_status; /**< status of bulk transfer */
 
-  int outfd; /**< I/Q output file descriptor */
-
   starcoder_queue* q; // Boost ring buffer for storing IQ data
 
   transfer_error_callback_func err_func; /**< error callback */
@@ -189,17 +187,7 @@ int ar2300_start_transfer(AR2300_HANDLE *ar2300);
 int ar2300_stop_transfer(AR2300_HANDLE *ar2300);
 
 /**
- * set the output file descriptor
- *
- * set the fd to which the output will be written to
- *
- * @param ar2300 handle to the ar2300 device
- * @param fd the file to write I/Q data to
- */
-void ar2300_set_fd(AR2300_HANDLE *ar2300, int fd);
-
-/**
- * set the ring buffer
+ * set the queue to write IQ packets to
  */
 void ar2300_set_q(AR2300_HANDLE *ar2300, starcoder_queue *q);
 
--- a/gr-starcoder/lib/ar2300_driver.h
+++ b/gr-starcoder/lib/ar2300_driver.h
@@ -115,7 +115,7 @@ typedef struct ar2300_handle {
 
   ar2300_bulk_status bulk_status; /**< status of bulk transfer */
 
-  starcoder_queue* q; // Boost ring buffer for storing IQ data
+  starcoder_queue* q; // Blocking queue for storing IQ data
 
   transfer_error_callback_func err_func; /**< error callback */
 
--- a/gr-starcoder/lib/ar2300_driver.h
+++ b/gr-starcoder/lib/ar2300_driver.h
@@ -23,7 +23,7 @@
 
 #define AR2300_USE_PTHREAD 1 /**< include threading code for event handler */
 #define AR2300_USE_SYSLOG 1  /**< output debug messages to syslog */
-#include "starcoder_queue.h"
+#include "blocking_queue.h"
 /*
  * constants related to receiver spec
  */
@@ -115,7 +115,7 @@ typedef struct ar2300_handle {
 
   ar2300_bulk_status bulk_status; /**< status of bulk transfer */
 
-  starcoder_queue* queue_; // Blocking queue for storing IQ data
+  blocking_queue* queue_; // Blocking queue for storing IQ data
 
   transfer_error_callback_func err_func; /**< error callback */
 
@@ -189,7 +189,7 @@ int ar2300_stop_transfer(AR2300_HANDLE *ar2300);
 /**
  * set the queue to write IQ packets to
  */
-void ar2300_set_queue(AR2300_HANDLE *ar2300, starcoder_queue *q);
+void ar2300_set_queue(AR2300_HANDLE *ar2300, blocking_queue *q);
 
 /**
  * start an event handling thread
--- a/gr-starcoder/lib/ar2300_driver.h
+++ b/gr-starcoder/lib/ar2300_driver.h
@@ -115,7 +115,7 @@ typedef struct ar2300_handle {
 
   ar2300_bulk_status bulk_status; /**< status of bulk transfer */
 
-  starcoder_queue* q; // Blocking queue for storing IQ data
+  starcoder_queue* queue_; // Blocking queue for storing IQ data
 
   transfer_error_callback_func err_func; /**< error callback */
 
@@ -189,7 +189,7 @@ int ar2300_stop_transfer(AR2300_HANDLE *ar2300);
 /**
  * set the queue to write IQ packets to
  */
-void ar2300_set_q(AR2300_HANDLE *ar2300, starcoder_queue *q);
+void ar2300_set_queue(AR2300_HANDLE *ar2300, starcoder_queue *q);
 
 /**
  * start an event handling thread
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -87,7 +87,7 @@ void ar2300_receiver::start() {
      throw std::runtime_error("ar2300_receiver::initialize");
   }
 
-  // Set the ring buffer for received data
+  // Set the blocking queue for received data
   ar2300_set_q(ar2300, &q);
 
   // Set the callback for error handling
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -87,9 +87,6 @@ void ar2300_receiver::start() {
      throw std::runtime_error("ar2300_receiver::initialize");
   }
 
-  // Set the pipe for writing received data
-  ar2300_set_fd(ar2300, read_pipe[PIPE_WRITE]);
-
   // Set the ring buffer for received data
   ar2300_set_q(ar2300, &q);
 
@@ -153,21 +150,4 @@ int ar2300_receiver::read(char* buf, int size) {
   int ret = q.pop(buf, size);
 
   return ret;
-}
-
-/*
- * Select data
- */
-int ar2300_receiver::select_for_read(int timeout) {
-  // Set timeout
-  struct timeval tv;
-  tv.tv_sec = (int)(timeout / 1000.0);
-  tv.tv_usec = (timeout % 1000) * 1000;
-
-  // Initialize file descriptor
-  fd_set rfds;
-  FD_ZERO(&rfds);
-  FD_SET(read_pipe[PIPE_READ], &rfds);
-
-  return ::select(FD_SETSIZE, &rfds, (fd_set*)NULL, (fd_set*)NULL, &tv);
 }
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -48,7 +48,7 @@ void err_callback(struct libusb_transfer* transfer, int code) {
  * Constructor
  */
 ar2300_receiver::ar2300_receiver(int buffer_size) :
-  q(buffer_size)
+  queue_(buffer_size)
 {
   context = NULL;
   ar2300 = NULL;
@@ -90,7 +90,7 @@ void ar2300_receiver::start() {
   }
 
   // Set the blocking queue for received data
-  ar2300_set_q(ar2300, &q);
+  ar2300_set_queue(ar2300, &queue_);
 
   // Set the callback for error handling
   ar2300_set_err_handler(ar2300, err_callback);
@@ -148,7 +148,7 @@ int ar2300_receiver::read(char* buf, int size) {
     throw std::runtime_error("ar2300_receiver::read");
   }
 
-  int ret = q.pop(buf, size);
+  int ret = queue_.pop(buf, size);
 
   return ret;
 }
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -47,7 +47,9 @@ void err_callback(struct libusb_transfer* transfer, int code) {
 /*
  * Constructor
  */
-ar2300_receiver::ar2300_receiver() {
+ar2300_receiver::ar2300_receiver(int buffer_size) :
+  q(buffer_size)
+{
   context = NULL;
   ar2300 = NULL;
   read_pipe[PIPE_READ] = -1;
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -136,8 +136,7 @@ void ar2300_receiver::stop() {
 
 /*
  * Read IQ data from device
- * @return: 0  - Select is timeout or there's nothing to read
- *          >0 - Number of bytes read
+ * @return: Number of bytes read
  */
 int ar2300_receiver::read(char* buf, int size) {
 
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -143,12 +143,12 @@ void ar2300_receiver::stop() {
  *          >0 - Number of bytes read
  */
 int ar2300_receiver::read(char* buf, int size, int timeout) {
-/*
+
   if (err_code != ERROR_CODE_NA) {
     stop();
     fprintf(stderr, "ar2300_receiver::read: something error occurred while reading data. err_code=%d\n", err_code);
     throw std::runtime_error("ar2300_receiver::read");
-  }*/
+  }
 
   int ret = q.pop(buf, size);
 
--- a/gr-starcoder/lib/ar2300_receiver.cc
+++ b/gr-starcoder/lib/ar2300_receiver.cc
@@ -90,6 +90,9 @@ void ar2300_receiver::start() {
   // Set the pipe for writing received data
   ar2300_set_fd(ar2300, read_pipe[PIPE_WRITE]);
 
+  // Set the ring buffer for received data
+  ar2300_set_q(ar2300, &q);
+
   // Set the callback for error handling
   ar2300_set_err_handler(ar2300, err_callback);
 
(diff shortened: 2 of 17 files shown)


gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+    q_.push(arr[i]);
+  }
+  if (was_empty) {
+    condition_var_.notify_one();
+  }
+  return size;
+}
+
+size_t starcoder_queue::pop(char *arr, size_t size) {
+  boost::mutex::scoped_lock lock(mutex_);
```
anuraaga: I think this prevents any more `push` while `pop` is waiting on the condition variable. Since we only have a single consumer, I think it's safe to have the lock after the wait.

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+  int i = 0;
+  while (i < size && !q_.empty()) {
+    arr[i] = q_.front();
+    q_.pop();
+    i ++;
+  }
+  return i;
+}
+
+size_t cpp_callback_push(starcoder_queue* q, const char *arr, size_t size) {
```
anuraaga: This isn't really used as a callback, I think just `pushToQueue` or similar semantic naming is fine.

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this software; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "starcoder_queue.h"
+
+starcoder_queue::starcoder_queue() :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(1048576)))
```
anuraaga: Pass the size as a constructor parameter.

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+  boost::mutex::scoped_lock lock(mutex_);
+  while (q_.empty()) {
+    condition_var_.wait(lock);
+  }
+
+  int i = 0;
+  while (i < size && !q_.empty()) {
+    arr[i] = q_.front();
+    q_.pop();
+    i ++;
```
anuraaga: `i++`

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+
+starcoder_queue::starcoder_queue() :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(1048576)))
+{ }
+
+size_t starcoder_queue::push(const char *arr, size_t size) {
+  boost::mutex::scoped_lock lock(mutex_);
+  bool const was_empty = q_.empty();
+  for (int i=0; i < size; i++) {
+    q_.push(arr[i]);
```
anuraaga: Is this not the right documentation? I see `push_back` and `pop_front` instead of `push`, `pop`.
https://www.boost.org/doc/libs/1_66_0/boost/circular_buffer/base.hpp

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "starcoder_queue.h"
+
+starcoder_queue::starcoder_queue() :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(1048576)))
+{ }
+
+size_t starcoder_queue::push(const char *arr, size_t size) {
+  boost::mutex::scoped_lock lock(mutex_);
+  bool const was_empty = q_.empty();
+  for (int i=0; i < size; i++) {
```
anuraaga: `i = 0`

gr-starcoder/lib/starcoder_queue.cc
reiinakano: The call to `condition_var_.wait(lock);` below will atomically unlock the mutex while waiting. I used this reference for this https://www.justsoftwaresolutions.co.uk/threading/implementing-a-thread-safe-queue-using-condition-variables.html

gr-starcoder/lib/starcoder_queue.cc
reiinakano: I'm using the `std::queue` as a container over `circular_buffer` http://en.cppreference.com/w/cpp/container/queue

gr-starcoder/lib/ar2300_driver.h
```diff
@@ -187,14 +187,9 @@ int ar2300_start_transfer(AR2300_HANDLE *ar2300);
...
- * set the output file descriptor
- *
- * set the fd to which the output will be written to
- *
- * @param ar2300 handle to the ar2300 device
- * @param fd the file to write I/Q data to
+ * set the queue to write IQ packets to
  */
-void ar2300_set_fd(AR2300_HANDLE *ar2300, int fd);
+void ar2300_set_q(AR2300_HANDLE *ar2300, starcoder_queue *q);
```
kenichi-fukushima: Not sure `q` is an accepted common abbreviation of `queue`. Spell out as `ar2300_set_queue`.

gr-starcoder/lib/ar2300_receiver.h
```diff
@@ -51,6 +52,9 @@ class ar2300_receiver {
   // Pipe for reading data from AR2300
   int read_pipe[2];
 
+  // Blocking queue
+  starcoder_queue q;
```
kenichi-fukushima: same as above

gr-starcoder/lib/ar2300_driver.h
```diff
@@ -115,7 +115,7 @@ typedef struct ar2300_handle {
 
   ar2300_bulk_status bulk_status; /**< status of bulk transfer */
 
-  int outfd; /**< I/Q output file descriptor */
+  starcoder_queue* q; // Blocking queue for storing IQ data
```
kenichi-fukushima: spell out `queue` as I explained below.

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,52 @@
...
+#include <boost/thread/mutex.hpp>
+  typedef std::queue<char, boost::circular_buffer<char>> internal_ring_buffer;
+  class starcoder_queue {
+  public:
+    starcoder_queue();
+    size_t push(const char*, size_t);
+    size_t pop(char*, size_t);
+  private:
+    internal_ring_buffer q_;
+    boost::fibers::condition_variable_any condition_var_;
```
kenichi-fukushima: use standard mutex and condition variable classes of C++11.
http://en.cppreference.com/w/cpp/thread/mutex
http://en.cppreference.com/w/cpp/thread/condition_variable

reiinakano:
Comments addressed

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include "starcoder_queue.h"
+
+starcoder_queue::starcoder_queue(int buffer_size) :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(buffer_size)))
+{ }
+
+size_t starcoder_queue::push(const char *arr, size_t size) {
+  std::unique_lock<std::mutex> lock(mutex_);
```
anuraaga: How about using `std::defer_lock`?

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,61 @@
...
+  return size;
+}
+
+size_t starcoder_queue::pop(char *arr, size_t size) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (q_.empty()) {
+    condition_var_.wait(lock);
+  }
+
+  int i = 0;
```
anuraaga: Let's name this `popped`

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "starcoder_queue.h"
+
+starcoder_queue::starcoder_queue(int buffer_size) :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(buffer_size)))
+{ }
+
+size_t starcoder_queue::push(const char *arr, size_t size) {
```
gavinreaney: If the circular buffer is full, I think this will overwrite the data at the start of the buffer. Is that what we want? I guess the previous code would block. Neither approach is very ideal but I think we'd want to know if this condition happens.

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this software; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "starcoder_queue.h"
+
+starcoder_queue::starcoder_queue(int buffer_size) :
+  q_(internal_ring_buffer(boost::circular_buffer<char>(buffer_size)))
```
kenichi-fukushima: (It was long before when I was actively using C++, so please disregard if my comment is pointless.)

Can you write 
`q_(boost::circular_buffer<char>(buffer_size))`
?

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,52 @@
...
+#include <mutex>
+#include <condition_variable>
+  typedef std::queue<char, boost::circular_buffer<char>> internal_ring_buffer;
+  class starcoder_queue {
+  public:
+    starcoder_queue(int buffer_size);
+    size_t push(const char*, size_t);
+    size_t pop(char*, size_t);
+  private:
+    internal_ring_buffer q_;
```
kenichi-fukushima: `queue_`. Use `q` only for short-lived local variables.

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,52 @@
...
+
+/* This header can be read by both C and C++ compilers */
+#ifndef STARCODER_QUEUE_H
+#define STARCODER_QUEUE_H
+#ifdef __cplusplus
+#include <queue>
+#include <boost/circular_buffer.hpp>
+#include <mutex>
+#include <condition_variable>
+  typedef std::queue<char, boost::circular_buffer<char>> internal_ring_buffer;
```
kenichi-fukushima: no indent this code block.

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,52 @@
...
+  };
+#else
+  typedef
+    struct starcoder_queue
+      starcoder_queue;
+#endif
+#ifdef __cplusplus
+extern "C" {
+#endif
+  extern size_t push_to_queue(starcoder_queue*, const char*, size_t);
```
kenichi-fukushima: no indent.  Preprocessor conditionals don't typically create indentation.

kenichi-fukushima: Move those two functions above the class definition. The public interface exposed by this file is the functions and the class is implementation detail.  A rule of thumb is public interface comes first and then implementation details follow in a file.

kenichi-fukushima: Do not omit argument names in forward declarations. They may play a great role of communicating purpose/meaning of those arguments. Without them, a user of this code has no idea what they should pass into those functions.

gr-starcoder/lib/ar2300_receiver.cc
```diff
@@ -136,49 +138,17 @@ void ar2300_receiver::stop() {
...
-    return 0;
-  }
-
-  int ret = ::read(read_pipe[PIPE_READ], buf, size);
-  if (ret < 0) {
-    stop();
-    fprintf(stderr, "ar2300_receiver::read: failed to read data from the device. ret=%d\n", ret);
-    throw std::runtime_error("ar2300_receiver::read");
-  }
+  int ret = queue_.pop(buf, size);
```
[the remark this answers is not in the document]
reiinakano: Good catch. I initially thought it would be fine, since https://wiki.gnuradio.org/index.php/BlocksCodingGuide says that condition variable `wait`s are fine to block indefinitely. When I tested it out though, I couldn't stop the flowgraph by simply calling `stop`. So no harm keeping the timeout.

gr-starcoder/lib/starcoder_queue.h
reiinakano: If I move it to the top, the compiler seems to error out since I haven't declared it yet.

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,59 @@
...
+{ }
+
+size_t starcoder_queue::push(const char *arr, size_t size) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  bool const was_empty = q_.empty();
+  for (int i = 0; i < size; i++) {
```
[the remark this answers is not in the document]
kenichi-fukushima: +1

gr-starcoder/lib/starcoder_queue.cc
```diff
@@ -0,0 +1,60 @@
...
+  if (was_empty) {
+    condition_var_.notify_one();
+  }
+  return size;
+}
+
+size_t starcoder_queue::pop(char *arr, size_t size, int timeout_ms) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (queue_.empty()) {
+    condition_var_.wait_for(lock, std::chrono::milliseconds(timeout_ms));
```
kenichi-fukushima: The documentation says the thread acquires lock when this function returns for whatever reasons including timeout, so it is safe to proceed. However, it is surprising to me that the code proceeds when the pre-condition is not met. Please either leave a comment that you intentionally proceed regardless of the return value, or check the return value and do early return when timeout.

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,51 @@
...
+#include <mutex>
+#include <condition_variable>
+#include <boost/lockfree/spsc_queue.hpp>
+class starcoder_queue {
+  public:
+    starcoder_queue(int buffer_size);
+    size_t push(const char*, size_t);
+    size_t pop(char*, size_t, int);
+  private:
+    boost::lockfree::spsc_queue<char> queue_;
```
kenichi-fukushima: Nice. `spsc_queue` is thread-safe so you can get rid of mutex and condvar.
https://www.boost.org/doc/libs/1_53_0/doc/html/boost/lockfree/spsc_queue.html#idp67134384-bb

Everything in boost::lockfree seems to be thread-safe.
https://www.boost.org/doc/libs/1_66_0/doc/html/lockfree.html

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,51 @@
...
+ * along with this software; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+/* This header can be read by both C and C++ compilers */
+#ifndef STARCODER_QUEUE_H
+#define STARCODER_QUEUE_H
+#ifdef __cplusplus
+#include <queue>
```
kenichi-fukushima: Remove this include.

gr-starcoder/lib/starcoder_queue.cc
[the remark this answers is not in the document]
reiinakano: I changed the implementation from `std::queue` to `spsc_queue`. This is already faster than the previous implementation (no precise benchmarks, but the CPU usage reported by system monitor went from 2-4% to 0-1%). The code is also much cleaner. I considered just `circular_buffer` instead of `spsc_queue` but I found this discussion: https://stackoverflow.com/questions/35675261/why-is-boostcircular-buffer-so-slow-in-my-benchmark

gr-starcoder/lib/starcoder_queue.cc
reiinakano: `spsc_queue` now won't write more than the available space left in the buffer

gr-starcoder/lib/starcoder_queue.h
reiinakano: Actually, the very first implementation of this was written with `spsc_queue`. I need this to be a blocking queue actually, since GNURadio will just call `pop` repeatedly and cause 100% CPU usage. So now the solution is combination of `spsc_queue` and the mutexes.

gr-starcoder/lib/starcoder_queue.cc
reiinakano: Oops. Good catch. This was a bug

gr-starcoder/lib/starcoder_queue.h
kenichi-fukushima: Ah, got it. Then, please leave a comment. Basically, any decisions made by the code author that are not obvious to other people is worth commenting.

reiinakano: Never mind, I can get rid of the mutex but keep the `cond_variable` I think. Let me do some more reading..

gr-starcoder/lib/starcoder_queue.h
```diff
@@ -0,0 +1,54 @@
...
+class starcoder_queue {
+  public:
+    starcoder_queue(int buffer_size);
+    size_t push(const char*, size_t);
+    size_t pop(char*, size_t, int);
+  private:
+    // Although we use the threadsafe spsc_queue here, we still need to keep the
+    // mutex for the sole purpose of waiting for the queue to be non-empty.
+    // This is since GNURadio will repeatedly call the non-blocking pop and
+    // consume 100% CPU.
```
anuraaga: I think more accurately this is because AR2300 samples at 1Mhz while our CPUs are at Ghzs, so there is a lot of blocking. Perhaps we will have faster sampling in the future where spin lock is more efficient than a condition var - do you think we should rename this so it's not so general as to preclude such a case? If not would at least document the sampling rate relationship

reiinakano: Yeah, the queue is almost always empty because of this. Anyway, by the time our sampling rates are comparable to our CPU rates, I doubt this code would still be relevant. 

So yeah maybe `starcoder_queue` is too general. What do you think about `blocking_queue`? Weird that Boost or the standard lib doesn't already have this.

anuraaga: Java has blocking queue but lock-free is a library called disruptor - I guess both have different philosphies ;)

`blocking_queue` SGTM

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -42,9 +42,9 @@ namespace gr {
       : gr::sync_block("ar2300_source",
               gr::io_signature::make(0, 0, 0),
               gr::io_signature::make(1, 1, sizeof(gr_complex))),
-        receiver(new ar2300_receiver())
+        receiver(new ar2300_receiver(10485760)), // 10MB buffer
+        timeout_ms(10000)
```
kenichi-fukushima: The previous value was 1000. Is this change intentional?

gr-starcoder/lib/blocking_queue.h
```diff
@@ -0,0 +1,54 @@
...
+};
+#else
+typedef
+  struct blocking_queue
+    blocking_queue;
+#endif
+#ifdef __cplusplus
+extern "C" {
+#endif
+extern size_t push_to_queue(blocking_queue* q, const char* arr, size_t size);
```
kenichi-fukushima: (nit/optional) A naming convention in C API design is to name a function after the main data structure that it operates on.  Because C doesn't have such concepts as namespace or package, this is the only way to avoid name collision and improve code readability. In this case, you could name this function `blocking_queue_push` and the below `blocking_queue_pop`.

gr-starcoder/lib/blocking_queue.cc
```diff
@@ -0,0 +1,56 @@
...
+
+blocking_queue::blocking_queue(int buffer_size) :
+  queue_(buffer_size)
+{ }
+
+size_t blocking_queue::push(const char *arr, size_t size) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  bool const was_empty = queue_.empty();
+  int pushed = queue_.push(arr, size);
+  if (was_empty) {
```
anuraaga: `pushed > 0 && was_empty`

gr-starcoder/lib/blocking_queue.cc
```diff
@@ -0,0 +1,56 @@
...
+
+#include "blocking_queue.h"
+#include <chrono>
+
+blocking_queue::blocking_queue(int buffer_size) :
+  queue_(buffer_size)
+{ }
+
+size_t blocking_queue::push(const char *arr, size_t size) {
+  std::unique_lock<std::mutex> lock(mutex_);
```
anuraaga: From my reading of this, I don't think we need this lock since the underlying queue is thread safe

http://en.cppreference.com/w/cpp/thread/condition_variable/notify_one

kenichi-fukushima: Sorry for overlooking this. The example code in the document release the lock before calling `notify_one`. It seems reasonable to me because the waiting thread should return from `wait` with the lock acquired. I'm not  very sure what happens if another thread has been holding the lock when it calls `notify`. The documentation says _The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock_

PR #31, opened 2018-04-24
STELLAR-1044 [FOLLOWUP] Replace OS pipe in AR2300 driver with blocking queue

reiinakano:
Followup PR based on @anuraaga 's comments

### DIFF
--- a/gr-starcoder/lib/CMakeLists.txt
+++ b/gr-starcoder/lib/CMakeLists.txt
@@ -28,7 +28,7 @@ list(APPEND starcoder_sources
     init.c
     driver.c
     firmware.c
-    blocking_queue.cc
+    blocking_spsc_queue.cc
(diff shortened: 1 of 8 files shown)


reiinakano:
@kenichi-fukushima continuing this discussion here
> With that said, isn't the lock still needed so calls of empty() and push() are atomic? It is still possible that this code calls empty() and get non-zero result, another thread pops, make the queue empty, attempts to pop more elements and waits, then this code pushes and return without calling notify_one().

I think since we will only have one producer and consumer, this situation won't happen.

reiinakano:
http://www.cplusplus.com/reference/condition_variable/condition_variable/notify_one/
>Unblocks one of the threads currently waiting for this condition. If no threads are waiting, the function does nothing.

PR #32, opened 2018-04-24
Add dockerfile which prepares a running starcoder image.

anuraaga:
The Dockerfile runs the build, including `gnuradio` prefix setup, and copies the prefix into the image. The default command of the image is starcoder.

- Also migrated usages of `go get` to more efficient manual download / `dep ensure`
- Disabled octoclock in `uhd` build for now - I feel like i have previously been able to build with it, but for some reason couldn't, and I feel like there is an obvious bug in their `CMakeLists.txt` so will debug it. Anyways, we don't use octoclock

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
           keys:
           - v1-dependencies
 
-      - run: ./gradlew install
+      - run: ./gradlew --no-daemon install
 
       - save_cache:
           paths:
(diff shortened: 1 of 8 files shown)


anuraaga:
`docker run -it --rm infostellarinc/starcoder`

reiinakano:
How do I pass in arguments to the `starcoder` call? Just appending them like `docker run -it --rm infostellarinc/starcoder --help` doesn't work.

anuraaga:
Guess I need to use entrypoint instead of command to allow that usage. Will push a new image tomorrow, though for viper flags you could also try env variables maybe

anuraaga:
Updated to use entrypoint, the docker image should work like you expect now

PR #33, opened 2018-04-25
STELLAR-1022 Rewrite enqueue_msg_sink block in C++

reiinakano:
Previously, it wasn't a big issue that this block was written in Python since packets don't arrive nearly as fast as samples. But since we're now sending the samples in blocks to send partial.bin files, we want to make this block as efficient as possible.

~~Also changed wiring in Starcoder to look for blocks with a `starcoder_observe` method. This method is called to retrieve the block's received messages.~~ Starcoder now looks for blocks with a `register_queue_pointer` method and drops in a pointer to the queue. It then waits for messages on this queue directly in C++


server/starcoder.go
```diff
@@ -474,47 +461,38 @@ func (s *Starcoder) startFlowGraph(modAndImport *moduleAndClassNames, request *p
...
+
+			iter := 0
+			key := python.PyString_FromString("")
+			defer key.DecRef()
+			val := python.PyString_FromString("")
+			defer val.DecRef()
+			for python.PyDict_Next(flowGraphDict, &iter, &key, &val) {
+				// Verify instance has starcoder_observe
+				hasStarcoderObserve := val.HasAttrString("starcoder_observe")
+				if hasStarcoderObserve == 1 {
```
anuraaga: I believe it should be relatively simple to access the C++ ptr instead of going through python, I pointed at some possible methods in a previous PR.

anuraaga: Until we get to swigging the gnuradio classes into golang for better direct access, you could have a C method that dequeues given a pointer to the block which should be trivial to call from golang

gr-starcoder/lib/enqueue_message_sink_impl.h
```diff
@@ -0,0 +1,55 @@
...
+     public:
+      enqueue_message_sink_impl();
+      ~enqueue_message_sink_impl();
+
+      // Where all the action really happens
+      int work(int noutput_items,
+         gr_vector_const_void_star &input_items,
+         gr_vector_void_star &output_items);
+
+      std::string starcoder_observe();
```
anuraaga: You might want to split out a class that is the queue and this method from the block. Then you could have a C++ / C library shared between the gnuradio block and starcoder

gr-starcoder/lib/enqueue_message_sink_impl.h
```diff
@@ -0,0 +1,55 @@
...
+
+    class enqueue_message_sink_impl : public enqueue_message_sink
+    {
+     private:
+      std::queue<std::string> queue_;
+      std::mutex mutex_;
+
+     public:
+      enqueue_message_sink_impl();
+      ~enqueue_message_sink_impl();
```
kenichi-fukushima: `virtual` if this classes used as an instance of `enqueue_mesage_sink` or one upper in the inheritance hierarchy. Calling a wrong destructor may end up with memory leak or any kind of bug.

gr-starcoder/include/starcoder/enqueue_message_sink.h
```diff
@@ -0,0 +1,60 @@
...
+     */
+    class STARCODER_API enqueue_message_sink : virtual public gr::sync_block
+    {
+     public:
+      typedef boost::shared_ptr<enqueue_message_sink> sptr;
+
+      /*!
+       * \brief Return a shared_ptr to a new instance of starcoder::enqueue_message_sink.
+       *
+       * To avoid accidental use of raw pointers, starcoder::enqueue_message_sink's
```
kenichi-fukushima: If you want to hide constructors you can just make them `private`. (you may have to also hide copy-constructor and move-constructor too)

kenichi-fukushima: Factory methods to control class instantiation is a universal concept seen in any language so you don't have to specifically mention it. Mention of implementation detail is rather considered an anti-pattern.

server/starcoder.go
reiinakano: While it's easy to access the C++ pointer from the Swig wrapper, I'm not sure how Go will know what to do with it.

I've read some `cgo` documentation, and it seems that we need to compile Go with knowledge of the C++ library so it knows what to do with a random C++ ptr. Also, we need to somehow wrap C++ with C since `cgo` only works with C. (Please correct me if I'm wrong since I'm not 100% sure about any of this.)

This is where I'm stuck, and it's also the reason why I haven't been able to implement a direct Go callback instead of this enqueue_message_sink solution yet. My plan was for this to be the last iteration of enqueue_message_sink (since we need the efficiency of C++), then the next would be to implement the more native go bindings, which I think is more complicated.

reiinakano: Didn't see your second comment. Let me think about it a bit.

gr-starcoder/include/starcoder/enqueue_message_sink.h
reiinakano: @kenichi-fukushima All your comments are actually from code auto-generated by a GNURadio tool for building C++ blocks.

kenichi-fukushima: Ah, ok. If the inheritance pattern is defined by GnuRadio, let's follow it.

I still recommend you delete useless junks such this comment after files are created. They bring confusion, but no value.

gr-starcoder/include/starcoder/enqueue_message_sink.h
```diff
@@ -0,0 +1,60 @@
+/* -*- c++ -*- */
```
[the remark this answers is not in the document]
reiinakano: I'm not really sure.. I tried removing it and it seems to work, but since this is auto-generated I don't want to remove it.

server/starcoder.go
reiinakano: I'm a bit stuck. There really doesn't seem to be a simple way to extract a valid C++ pointer from a Python-swigged proxy. Even if I can get the pointer in integer form from SWIG via `int(swigProxy.this)`, the SWIG docs say:

>As much as you might be inclined to modify a pointer value directly from Python, don't. The hexadecimal encoding is not necessarily the same as the logical memory address of the underlying object. Instead it is the raw byte encoding of the pointer value. The encoding will vary depending on the native byte-ordering of the platform (i.e., big-endian vs. little-endian).

So I'm assuming I can't just pass that particular integer to `unsafe.Pointer` and expect it to work. And even if I could, the Go docs say:

> A uintptr is an integer, not a reference. Converting a Pointer to a uintptr creates an integer value with no pointer semantics. Even if a uintptr holds the address of some object, the garbage collector will not update that uintptr's value if the object moves, nor will that uintptr keep the object from being reclaimed.

The closest thing I could see was this: https://stackoverflow.com/questions/28904033/extracting-swig-wrapped-c-instance-pointer-for-use-in-cython but I'm unsure how to do something like this in Go. 

Do you have any ideas of what the best way is to achieve this? I read https://stackoverflow.com/questions/15602280/convert-pointer-to-int-and-back-to-typed-object, and I was thinking of maybe just passing an integer (typecast from *enqueue_message_sink) directly to a C function and re-converting that to a pointer inside, but seems dangerous.

reiinakano: To clarify: 
`int(swigProxy)` 
-> 
```c
c_function(x int) string {
//pass (enqueue_message_sink *)(x) to C++ function
}
```
should just work?

Do you think there's a need to define a function
```c++
int get_ptr() {
  return reinterpret_cast<unsigned long>(this);
}
```
to get the integer or is `int(swigProxy)` enough?

anuraaga: Ah yeah think using `reinterpret_cast` in your own helper will be better than using `int(swigProxy)` I'm not sure exactly what it means by "raw byte encoding of pointer"

reiinakano:
I've addressed the comments. Thanks

server/starcoder.go
```diff
@@ -474,47 +462,47 @@ func (s *Starcoder) startFlowGraph(modAndImport *moduleAndClassNames, request *p
...
+				// Verify instance has register_queue_pointer
+				hasStarcoderObserve := val.HasAttrString("register_queue_pointer")
+				if hasStarcoderObserve == 1 {
+					k := python.PyString_AsString(key)
+					newQ := cqueue.NewCQueue()
+					pyQPtr := python.PyLong_FromUnsignedLongLong(newQ.GetPtr())
+					result := val.CallMethodObjArgs("register_queue_pointer", pyQPtr)
+					pyQPtr.DecRef()
+					if result == nil {
+						err = errors.New(getExceptionString())
```
anuraaga: `newQ.Delete()`

cqueue/c_queue.cc
```diff
@@ -0,0 +1,45 @@
...
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "c_queue.h"
+
+c_queue::c_queue() {}
+
+void c_queue::push(std::string item) {
+  std::unique_lock<std::mutex> lock(mutex_);
```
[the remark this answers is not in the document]
anuraaga: Also if you make it blocking, I think you can remove the polling on the golang side just loop with blocking pops which should have lower latency than polling (guess we have a recently created very convenient class for this :) )

cqueue/c_queue.h
```diff
@@ -0,0 +1,34 @@
...
+#define C_QUEUE_H
+#include <mutex>
+#include <string>
+#include <queue>
+class c_queue {
+  public:
+    c_queue();
+    void push(std::string);
+    std::string pop();
+    unsigned long get_ptr();
```
kenichi-fukushima: `uintptr_t get_ptr() const;`

cqueue/c_queue.h
```diff
@@ -0,0 +1,34 @@
...
+  private:
+    std::queue<std::string> queue_;
+    std::mutex mutex_;
+};
+c_queue *c_queue_from_ptr(unsigned long ptr);
```
kenichi-fukushima: `uintptr_t`

cqueue/c_queue.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef C_QUEUE_H
+#define C_QUEUE_H
+#include <mutex>
+#include <string>
+#include <queue>
+class c_queue {
```
kenichi-fukushima: what is `c` in the name? Something like `string_queue` sounds more reasonable to me.

cqueue/c_queue.h
```diff
@@ -0,0 +1,34 @@
...
+  public:
+    c_queue();
+    void push(std::string);
+    std::string pop();
```
kenichi-fukushima: Just FYI: a function to remove an element from collection and return it is exception-unsafe in general. But in this case it is ok because the element type is string. Its copy constructor throws only when the system has run out of memory.
https://accu.org/index.php/journals/444
http://www.cplusplus.com/reference/string/string/string/

cqueue/c_queue.h
kenichi-fukushima: Understood the context. Then, maybe you can use `uint64_t`.

cqueue/c_queue.h
reiinakano: Thanks for the reference! I added it in the code as a comment.

anuraaga:
boost unfortunately still doesn't setup pkg-config. Can you add `environment 'C_INCLUDE_PAATH', "${gnuradioRoot}/include"?

cqueue/c_queue.go
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package cqueue
+
+// Thin wrapper around the c_queue C++ class so
+// we only need to deal with unrecognized variables in this file.
+type CQueue struct {
```
anuraaga: `CStringQueue`?

cqueue/string_queue.cc
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+std::string string_queue::pop() {
+  std::string a;
+  queue_.pop(a);
+  return a;
+}
+
+std::string string_queue::block_pop() {
```
anuraaga: `blocking_pop` or possibly `take` which is the Java convention.

reiinakano:
I added the line like this:

```groovy
    // For some reason, cmake has trouble finding boost in certain situations.
    environment 'BOOST_ROOT', gnuradioRoot

    environment 'C_INCLUDE_PATH', "${gnuradioRoot}/include"

    environment 'PKG_CONFIG_PATH', "${gnuradioRoot}/lib/pkgconfig"
```

but it still doesn't work

kenichi-fukushima:
@reiinakano  Did you push PoC code to this branch unintentionally?

reiinakano:
Whoops, anyway I can roll it back. What is your opinion on the separate class?

cqueue/string_queue.h
```diff
@@ -0,0 +1,41 @@
...
+    std::string pop();
+    std::string blocking_pop();
+    unsigned long get_ptr();
+    void wake();
+  private:
+    boost::lockfree::spsc_queue<std::string> queue_;
+    std::condition_variable condition_var_;
+    std::mutex mutex_;
+};
+string_queue *string_queue_from_ptr(uint64_t ptr);
```
kenichi-fukushima: Move this function to the class's scope.

```
class string_queue {
  static string_queue *from_ptr(uint64_t ptr);
}
```

build.gradle
```diff
@@ -309,6 +309,8 @@ task setupPrefix(type: Exec) {
     // For some reason, cmake has trouble finding boost in certain situations.
     environment 'BOOST_ROOT', gnuradioRoot
 
+    environment 'C_INCLUDE_PATH', "${gnuradioRoot}/include"
```
anuraaga: Sorry should have clarified - add it to the `gobuild` task not here

cqueue/string_queue.h
```diff
@@ -0,0 +1,41 @@
...
+class string_queue {
+  public:
+    string_queue(int buffer_size);
+    void push(const std::string &str);
+    // This form of pop function is exception-unsafe, but is okay since we are returning
+    // a std::string. Its copy constructor throws only when the system has run out
+    // of memory. https://accu.org/index.php/journals/444
+    std::string pop();
+    std::string blocking_pop();
+    unsigned long get_ptr();
```
kenichi-fukushima: this function does not mutate the object so make it const.

```
```

cqueue/string_queue.h
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef STRING_QUEUE_H
+#define STRING_QUEUE_H
+#include <mutex>
+#include <condition_variable>
+#include <string>
+#include <boost/lockfree/spsc_queue.hpp>
+class string_queue {
```
kenichi-fukushima: I think semantics of this queue has diverged a lot from that of usual concurrent queues. A concurrent queue does not usually provide non-blocking pop with no result, active wake up from external code. It should have been written for a very specific application. Please comment about motivations for the unique semantics and how to use it.

cqueue/string_queue.cc
```diff
@@ -0,0 +1,53 @@
...
+  std::string a;
+  queue_.pop(a);
+  return a;
+}
+
+std::string string_queue::blocking_pop() {
+  std::string a;
+  std::unique_lock<std::mutex> lock(mutex_);
+  if (queue_.empty())
+    condition_var_.wait(lock);
```
gavinreaney: Note that it seems possible for string_queue::push to start execution after the queue.empty returns true but complete before condition_var_.wait is called. This will leave us in a state where the consumer is waiting but the queue is not empty. Does this matter?

reiinakano: It doesn't matter with how we use it, since once a flowgraph ends we call `wake()` on the queue and `pop` out everything that's left. But I guess this is unclear if someone else decides to use this class for an unrelated purpose (although not sure what that would be). I could maybe document this in a comment or put a lock around `push`. Wdyt?

reiinakano: Actually, I think it's better to lock around `push`. We should make sure to return any pushed messages as soon as it is pushed. Thanks for catching this. I'll fix it in a followup.

PR #34, opened 2018-05-01
Send back waterfall plot on a string_queue

reiinakano:
Adds `register_queue_pointer` method to waterfall plotter and sends back final waterfall plot via `string_queue` (if registered).

Also adds a lock around `string_queue.push()` as per https://github.com/infostellarinc/starcoder/pull/33#discussion_r185179199

### DIFF
--- a/cqueue/string_queue.cc
+++ b/cqueue/string_queue.cc
@@ -22,7 +22,9 @@
 string_queue::string_queue(int buffer_size) : queue_(buffer_size) {}
 
 void string_queue::push(const std::string &item) {
+  std::unique_lock<std::mutex> lock(mutex_);
   bool pushed = queue_.push(item);
+  lock.unlock();
   if (pushed) {
     condition_var_.notify_one();
   }
--- a/cqueue/string_queue.cc
+++ b/cqueue/string_queue.cc
@@ -19,7 +19,7 @@
 
 #include "string_queue.h"
 
-string_queue::string_queue(int buffer_size) : queue_(buffer_size) {}
+string_queue::string_queue(int buffer_size) : queue_(buffer_size), closed_(false) {}
 
 void string_queue::push(const std::string &item) {
   std::unique_lock<std::mutex> lock(mutex_);
@@ -39,14 +39,16 @@ std::string string_queue::pop() {
 std::string string_queue::blocking_pop() {
   std::string a;
   std::unique_lock<std::mutex> lock(mutex_);
-  if (queue_.empty())
+  while (queue_.empty() && !closed_)
     condition_var_.wait(lock);
-  // If woken up spuriously, will return an empty string.
   queue_.pop(a);
   return a;
 }
 
 void string_queue::wake() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  closed_ = true;
+  lock.unlock();
   condition_var_.notify_one();
 }
 
--- a/cqueue/string_queue.cc
+++ b/cqueue/string_queue.cc
@@ -39,8 +39,7 @@ std::string string_queue::pop() {
 std::string string_queue::blocking_pop() {
   std::string a;
   std::unique_lock<std::mutex> lock(mutex_);
-  while (queue_.empty() && !closed_)
-    condition_var_.wait(lock);
+  condition_var_.wait(lock, [this]{return (!queue_.empty() || closed_);});
   queue_.pop(a);
   return a;
 }
--- a/cqueue/string_queue.cc
+++ b/cqueue/string_queue.cc
@@ -45,7 +45,7 @@ std::string string_queue::blocking_pop() {
   return a;
 }
 
-void string_queue::wake() {
+void string_queue::close() {
   std::unique_lock<std::mutex> lock(mutex_);
   closed_ = true;
   lock.unlock();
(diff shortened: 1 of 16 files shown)


gr-starcoder/include/starcoder/waterfall_plotter.h
```diff
@@ -56,17 +56,10 @@ namespace gr {
...
-       *
-       * To avoid accidental use of raw pointers, starcoder::waterfall_plotter's
-       * constructor is in a private implementation
-       * class. starcoder::waterfall_plotter::make is the public interface for
-       * creating new instances.
-       */
       static sptr
       make(double samp_rate, double center_freq,
            int rps, size_t fft_size, char* filename);
+      virtual void register_queue_pointer(uint64_t ptr) = 0;
```
kenichi-fukushima: Just `register_queue` or `register_<some descriptive term>_queue`. Passing an object by pointer is very common so `pointer` is not needed.

gr-starcoder/lib/waterfall_plotter_impl.cc
```diff
@@ -171,12 +172,14 @@ namespace gr {
...
-      image_buffer = PyString_AsString(result);
-      if (image_buffer == NULL)
-        goto error;
+      if (string_queue_ != NULL) {
+        Py_ssize_t image_size = PyString_Size(result);
+        char *image_buffer = PyString_AsString(result);
+        if (image_buffer == NULL)
+          goto error;
+        const std::string image_binary(image_buffer, image_size);
+        string_queue_->push(image_binary);
```
kenichi-fukushima: I'm not sure how this code has to be performant but you may use the move semantics here to avoid copying if it has significant impact on performance. See the secand overload variant of `queue::push``
http://www.cplusplus.com/reference/queue/queue/push/

cqueue/string_queue.cc
```diff
@@ -22,7 +22,9 @@
 string_queue::string_queue(int buffer_size) : queue_(buffer_size) {}
 
 void string_queue::push(const std::string &item) {
+  std::unique_lock<std::mutex> lock(mutex_);
```
anuraaga: Is the reason we need this lock because we don't loop in `pop`? If so, I think we can avoid the lock and instead have `pop` check a separate boolean like `closed` to return out when we're done.

anuraaga: I think we avoid that problem in our other queue because `pop` has a loop. I think we can have a loop here too if you have a `bool closed` so the loop knows when to end.

reiinakano: Actually, that loop allows us to handle spurious wakes.

In fact, now that you mention it, looks like *that* queue has this behavior as well. It's just not as severe since we have a timeout + the GNURadio scheduler calls `pop` again very quickly after the timeout.

cqueue/string_queue.cc
```diff
@@ -22,7 +22,9 @@
...
   bool pushed = queue_.push(item);
+  lock.unlock();
```
gavinreaney: I think the unlock would need to happen after you call notify.

But let me think about whether we can use the (volatile) variable plus loop instead.

reiinakano: Hmm, I can't think of a scenario where we need the unlock after notify. Could you describe what you have in mind?

gavinreaney: Ah sorry, you are correct :)

cqueue/string_queue.cc
reiinakano: ping

anuraaga: I was thinking of changing `pop` to

```go
while(queue.empty() && !closed) {
  wait();
}
```
but yeah actually it doesn't change the need for the lock. It might still be better to do it that way though to avoid having to return to `go` and call back into `c` on spurious wakeups.

reiinakano: Done

cqueue/string_queue.cc
```diff
@@ -37,14 +39,16 @@ std::string string_queue::pop() {
...
+  while (queue_.empty() && !closed_)
     condition_var_.wait(lock);
-  // If woken up spuriously, will return an empty string.
   queue_.pop(a);
   return a;
 }
 
-void string_queue::wake() {
+void string_queue::close() {
+  std::unique_lock<std::mutex> lock(mutex_);
```
anuraaga: Since it's volatile don't need to lock

reiinakano: https://stackoverflow.com/a/4558031/5636655 says that `volatile` shouldn't be used for synchronization. Wdyt?

anuraaga: Oh didn't know about c++ `volatile` being weird like that, how about using `atomic` then?

anuraaga: Ah that's true - how about we use the version of `.wait` that takes a predicate? Looks like we don't have to implement the lock ourselves.

reiinakano: Done

cqueue/string_queue.h
```diff
@@ -37,10 +40,11 @@ class string_queue {
     std::string pop();
     std::string blocking_pop();
     unsigned long get_ptr() const;
-    void wake();
+    void close();
   private:
     boost::lockfree::spsc_queue<std::string> queue_;
     std::condition_variable condition_var_;
     std::mutex mutex_;
+    volatile bool closed_;
```
anuraaga: Since we're using a lock in the end, don't think this needs to be `volatile`

PR #36, opened 2018-05-07
Add a mock pygtk during installation to allow grcc install script to …

anuraaga:
…run.

GRCC installation does some extra steps like generating XML blocks and copying them over, so it's best to run it's cmake installation rather than trying to copy it manually. To avoid requiring GTK to build, we trick the build into thinking `pygtk` is installed.


anuraaga:
Yeah I'm building it (another 15 min or so...).

anuraaga:
Pushed the new image

PR #37, opened 2018-05-08
Don't delete the GTK mock after setup since it is still needed when r…

anuraaga:
…unning GRCC even for non-gui graphs.

Also don't set the grc blocks env variable since it doesn't seem to be needed.


anuraaga:
Related: https://github.com/gnuradio/gnuradio/pull/1766

PR #38, opened 2018-05-09
Add Waterfall-only flowgraph

reiinakano:
This PR introduces `waterfall_only.grc`, containing only AR2300 -> Decimating FIR Filter -> Waterfall Sink and partial.bin generator.

screenshot from 2018-05-09 12-30-38

____________
It also contains two bugfixes for the AR2300 source block:

The expected byte pattern for the AR2300 packets is as follows:
```
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
```

One bug occurs when a byte gets lost, and we lose alignment. This bug, as far as I've seen, only happens a few hundred samples into the beginning of a flow graph. e.g.

```
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxx0 xxxxxxxx
xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0 xxxxxxxx
```
The solution was to realign the bytes when this happened.

The second bug occurs more frequently, when we receive bytes that look like this:

```
xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx1 xxxxxxxx xxxxxxx0 xxxxxxxx xxxxxxx0
```

In this case, we just threw away the first two bytes and the last six, effectively losing a sample each time the `work` function was called, which for the `waterfall_only.grc` graph was once every 300-4000 samples. The solution was to just store any extra unused bytes at the end and reuse them in the next call.
_____________
Finally, I call `buf.close()` and `plt.close()` at the end of the plotting script, which was causing a pretty big memory leak.


gr-starcoder/python/plot_waterfall.py
```diff
@@ -111,4 +111,8 @@ def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
 
     buf = io.BytesIO()
```
anuraaga: I guess this can be

```python
with io.BytesIO() as buf:
  ...
```

without closing

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -81,6 +81,23 @@ namespace gr {
...
+
+      if (num_leftover_ > 0) {
+        for (int i = 0; i < num_leftover_; i++) {
+          sample[i] = leftover_[i];
+        }
+        for (int i = 0; i < 8-num_leftover_; i++) {
+          sample[i + num_leftover_] = in[i];
+        }
+        if (!validate_sample(sample)) {
+          GR_LOG_WARN(d_logger, boost::format("Reconstructed sample invalid"));
```
anuraaga: Does this happen due to data loss? Will we have a sense of AR2300 data loss by looking at the warning messages?

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -81,6 +81,23 @@ namespace gr {
                               const int inSize,
                               gr_complex* out)
     {
+      char sample[8];
+      int  out_index = 0;
```
anuraaga: Always one space after type

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -81,6 +81,23 @@ namespace gr {
...
+
+      if (num_leftover_ > 0) {
+        for (int i = 0; i < num_leftover_; i++) {
+          sample[i] = leftover_[i];
+        }
+        for (int i = 0; i < 8-num_leftover_; i++) {
```
anuraaga: In C++, we put spaaces in e.g. `8 - num_leftover_`

gr-starcoder/lib/ar2300_source_impl.cc
reiinakano: Well, this warning is only emitted if the reconstructed sample (from the previous work function) is corrupted. This would indicate we're missing at least 1-3 bytes in between samples. But since this sample is only reconstructed every 300-4k samples (per `work()` call), I don't think it will be indicative of data loss. So far, I have never seen this warning.

Also, this warning only detects corrupted samples. I don't actually know if data loss will happen in the form of corrupted samples. If data loss happens by the AR2300 just dropping the entire sample (8 bytes), then we have no way to detect it from this side.

gr-starcoder/python/plot_waterfall.py
```diff
@@ -109,6 +109,10 @@ def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
     if filename != '':
         plt.savefig(filename, bbox_inches='tight', pad_inches=0.2, format='png')
 
-    buf = io.BytesIO()
-    plt.savefig(buf, bbox_inches='tight', pad_inches=0.2, format='png')
-    return buf.getvalue()
+    with io.BytesIO() as buf:
+        buf = io.BytesIO()
+        plt.savefig(buf, bbox_inches='tight', pad_inches=0.2, format='png')
+        out = buf.getvalue()
```
anuraaga: I think you can just do
```python
plt.savefig()
plt.close()
return buf.getvalue()
```

cqueue/string_queue.cc
```diff
@@ -40,9 +40,7 @@ std::string string_queue::pop() {
 std::string string_queue::blocking_pop() {
   std::string a;
   std::unique_lock<std::mutex> lock(mutex_);
-  condition_var_.wait(lock, [this] {
-    return (!queue_.empty() || closed_);
-  });
+  condition_var_.wait(lock, [this] { return (!queue_.empty() || closed_); });
```
anuraaga: Can you run `clangFormat` again? Not sure why but it looks like the merge reverted all the changes instead of accepting them.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -68,6 +68,23 @@ int ar2300_source_impl::work(int n_output_items,
 
 int ar2300_source_impl::encode_ar2300(const char *in, const int inSize,
                                       gr_complex *out) {
+  char sample[8];
+  int out_index = 0;
+
+  if (num_leftover_ > 0) {
+    for (int i = 0; i < num_leftover_; i++) {
```
gavinreaney: This loop and the next could be replaced with the following, I think:
memcpy(sample, leftover_, num_leftover_)
memcpy(sample + num_leftover_, in, 8 - num_leftover_)

Or use std::copy to look more modern.

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -68,6 +68,23 @@ int ar2300_source_impl::work(int n_output_items,
...
+    for (int i = 0; i < 8 - num_leftover_; i++) {
+      sample[i + num_leftover_] = in[i];
+    }
+    if (!validate_sample(sample)) {
+      GR_LOG_WARN(d_logger, boost::format("Reconstructed sample invalid"));
+    } else {
+      out[out_index++] = parse_sample(sample);
+    }
+  }
+
```
gavinreaney: In the above loop, we may copy some of the data from 'in' to 'sample'. Should we skip past these when looking for the 'i_value' below?

gr-starcoder/lib/ar2300_source_impl.cc
```diff
@@ -108,6 +124,11 @@ int ar2300_source_impl::encode_ar2300(const char *in, const int inSize,
     }
   }
 
+  for (int i = 0; i < sample_index; i++) {
```
gavinreaney: Again, consider using memcpy or std::copy

gr-starcoder/lib/ar2300_source_impl.cc
reiinakano: Although I've never seen it happen, the reconstructed sample above might be invalid, and some of the first bytes we copied over from `in` above might be part of the next actual sample. Otherwise, the loop will just skip over them normally as it is right now. 

But I guess it would be easy to change the starting value of the loop depending on if the reconstructed sample is valid or not. Wdyt?

gavinreaney: If you are pretty sure it's harmless to keep it as-is then maybe that's fine (you could add a comment about it though).

This did make me think about another case though: if something below exits early because of an error case, should we reset num_leftover_ to zero? Or maybe it doesn't matter because the data in leftover_ will always be detected as invalid if that happens?

reiinakano: I changed the design a bit. It's simpler if we don't look *at all* for the `i_value` since the next one should be right after the reconstructed sample. And if it's corrupted, we already adjust the offset automatically below.

For the second case you mention, I don't think it's necessary to reset `num_leftover_` to zero for the reason you mentioned.

PR #40, opened 2018-05-09
Add / apply clang-format

anuraaga:


anuraaga:
Note the one manual change I made to add an equals sign. Without it, the formatting was very strange, and I guess it's standard to use equals with an initializer like that.

https://github.com/infostellarinc/starcoder/pull/40/commits/ca4703d3e060e54362930d68b05947c7cbed8040

PR #43, opened 2018-05-18
STELLAR-1143 Wire up Starcoder to receive packets

reiinakano:
This PR gives Starcoder clients the ability to send information *to* a GNURadio flowgraph.

GNURadio flowgraphs use a data structure called PMT (Polymorphic Data Type) that can contain arbitrary messages. These arbitrary messages can contain packets for transmission, commands to shift frequencies for Doppler correction, etc. 

This PMT representation has been converted to an equivalent PMT message in starcoder.proto so they can be constructed client-side and sent through gRPC. See https://gnuradio.org/doc/doxygen/page_pmt.html for documentation about PMTs.

A custom block `command_source` is also created to receive commands from Starcoder directly through the C++ `string_queue` and pass the message to whichever block it is connected to.

The data flow is as follows:
1. Client starts bidirectional stream, with the first packet containing flowgraph name to start and flowgraph starting parameters. This is the same as before.
2. Further packets from the client will contain the `Command` message, and this will contain the block ID to send the PMT to + the actual PMT.
3. Starcoder receives the command and sends the serialized gRPC `PMT` to the `command_source` targeted.
4. `command_source` deserializes the `PMT` and converts it to the GNURadio native `pmt_t` type.
5. PMT is sent to whichever block `command_source` is connected to.

In the future,`enqueue_msg_sink` will probably also convert the PMT received to gRPC format before sending it back up to Starcoder.


api/starcoder.proto
```diff
@@ -39,6 +39,166 @@ message RunFlowgraphRequest {
 
   // Blocks to stream back in the response
   repeated string block_id = 3;
+
+  // PMT command to send to a block
```
anuraaga: Let's group the different types of requests to the starcoder (gRPC streaming makes these sort of concepts a bit unnatural but we'll live with it), so maybe something like

```proto
message StarcoderRequest {

  oneof Request {
    RunFlowgraphRequest runFlowgraphRequest = 1;
    SendCommandRequest sendCommandRequest = 2;
  }
}

message RunFlowgraphRequest {
   flowgraph parameters...
}

message SendCommandRequest {
  command parameters...
}
```

api/starcoder.proto
```diff
@@ -39,6 +39,166 @@ message RunFlowgraphRequest {
...
+// Starcoder block for sending PMTs.
+message Command {
+  // Block ID to send the command message to.
+  string block_id = 1;
+
+  PMT pmt = 2;
+}
+
+// Protobuf format of GNURadio PMT (Polymorphic Message Type)
+message PMT {
```
anuraaga: Do we need the word `PMT`? Would a flowgraph have the words `pmt_pair`, etc in it? Unless this corresponds 1-1 with syntax in a flowgraph, we can avoid the word `pmt` since this is a different format.

api/starcoder.proto
```diff
@@ -39,6 +39,166 @@ message RunFlowgraphRequest {
...
+    PMTF64Vector f64_value = 10;
+
+    PMTC32Vector c32_value = 11;
+
+    PMTC64Vector c64_value = 12;
+  }
+}
+
+// Protobuf format of a PMT Uint8 Vector
+message PMTU8Vector {
```
anuraaga: Don't think we need these wrappers.

api/starcoder.proto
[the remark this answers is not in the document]
reiinakano: The Starcoder client generates PMT values (Starpass).

api/starcoder.proto
anuraaga: Don't you differentiate based on what field in `PMTUniformVector` is populated?

reiinakano: Oh! Yeah, you're right we don't need these wrappers :)

reiinakano:
Closing this first. Will open a PR with just the API change then merge the approved API back in here for later.

api/starcoder.proto
reiinakano: Looks like a `oneof` can't be repeated :(

anuraaga: Ugh :( Ok let's leave the wrapping then.

gr-starcoder/lib/proto_to_pmt.cc
```diff
@@ -0,0 +1,256 @@
...
+          grpc_pmt_uniform_vector.c64_value().value().size(), vec);
+    }
+    default:
+      throw("Type of uniform vector not set");
+  }
+}
+
+pmt::pmt_t convert_pmt_dict(const starcoder::Dict &grpc_pmt_dict) {
+  pmt::pmt_t dict = pmt::make_dict();
+  for (int i = 0; i < grpc_pmt_dict.entries_size(); i++) {
```
kenichi-fukushima: Doesn't below work?
`for (int i : grpc_pmt_dist.entries()) {`

api/starcoder.proto
```diff
@@ -100,7 +100,7 @@ message Pair {
 // Represents a PMT Tuple or Vector. A Tuple is an immutable list of PMTs.
 // It can only hold 10 elements. A Vector is a mutable list of PMTs.
 message List {
-  repeated BlockMessage values = 1;
+  repeated BlockMessage value = 1;
```
kenichi-fukushima: Maybe rename `entries` in `Dict` too?

cqueue/string_queue.h
```diff
@@ -41,6 +41,8 @@ class string_queue {
     std::string blocking_pop();
     unsigned long get_ptr() const;
     void close();
+    // TODO: Make this uint64_t
```
kenichi-fukushima: Why can't it happen now?

reiinakano:
Okay, I've integrated the changes. PTAL. @anuraaga, the previous no-cache build was cancelled so will need to rebuild it. I don't have write permissions to trigger a build.

gr-starcoder/lib/command_source_impl.h
```diff
@@ -0,0 +1,57 @@
...
+#include <boost/thread/thread.hpp>
+#include "string_queue.h"
+
+namespace gr {
+  namespace starcoder {
+
+    class command_source_impl : public command_source
+    {
+     private:
+      boost::thread *thread_;
```
kenichi-fukushima: Use C++11's standard `thread` class.
http://en.cppreference.com/w/cpp/thread/thread

But I would expect Gnuradio provide building blocks for parallel processing. Doesn't it?

api/starcoder.proto
reiinakano: Rename it to `entry`? I'm not sure what the convention is for naming `repeated` fields. Is it singular or plural?

cqueue/string_queue.h
reiinakano: Swig doesn't support direct conversion from Go `uint64` to `uint64_t`. Will need a bit more work which I think could be in a separate PR.

gr-starcoder/lib/command_source_impl.h
reiinakano: Actually, they just use `boost::thread` too.

gr-starcoder/lib/proto_to_pmt.cc
```diff
@@ -20,237 +20,235 @@
...
-}
-
-pmt::pmt_t convert_pmt_vector(const starcoder::List &grpc_pmt_vector) {
-  pmt::pmt_t vec =
-      pmt::make_vector(grpc_pmt_vector.value_size(), pmt::get_PMT_NIL());
-  for (int i = 0; i < grpc_pmt_vector.value_size(); i++) {
-    pmt::vector_set(vec, i, convert_pmt_proto(grpc_pmt_vector.value(i)));
+pmt::pmt_t convert_pmt_list(const starcoder::List &proto_pmt_list) {
+  int size = proto_pmt_list.value_size();
+  if (proto_pmt_list.type() == starcoder::List::TUPLE)
```
kenichi-fukushima: I recommend you don't omit curly braces around conditional statement body unless it's short enough to fit within one line.

fyi:
https://google.github.io/styleguide/cppguide.html#Conditionals

api/starcoder.proto
kenichi-fukushima: Our convention is to use singular for repeated fields.

reiinakano:
Done

api/starcoder.proto
reiinakano: Done

PR #44, opened 2018-05-21
[STELLAR-1143] Update API for transmission

reiinakano:
Change in gRPC API for supporting commands. View https://github.com/infostellarinc/starcoder/pull/43 for context.

### DIFF
--- a/api/starcoder.proto
+++ b/api/starcoder.proto
@@ -21,7 +21,18 @@ package starcoder;
 
 option go_package = "/api";
 
+// Request to run a flowgraph. Can be StartFlowgraphRequest for flowgraph
+// starting parameters or SendCommandRequest to send commands dynamically
+// while flowgraph is running.
 message RunFlowgraphRequest {
+  oneof Request {
+    StartFlowgraphRequest startFlowgraphRequest = 1;
+    SendCommandRequest sendCommandRequest = 2;
+  }
+}
+
+// Request to start a flowgraph. This is the first request sent in a stream.
+message StartFlowgraphRequest {
   // Filename of GNURadio script to run
   string filename = 1;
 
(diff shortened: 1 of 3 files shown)


api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
...
+// Command message sent to the flowgraph. This only works with the custom
+// Starcoder block for sending PMTs.
+message SendCommandRequest {
+  // Block ID to send the command message to.
+  string block_id = 1;
+
+  PMT pmt = 2;
+}
+
+// Protobuf format of GNURadio PMT (Polymorphic Message Type)
```
anuraaga: Let's name it `PmtMessage`.

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
   repeated string block_id = 3;
 }
 
+// Command message sent to the flowgraph. This only works with the custom
+// Starcoder block for sending PMTs.
+message SendCommandRequest {
+  // Block ID to send the command message to.
+  string block_id = 1;
+
+  PMT pmt = 2;
```
anuraaga: `// Will be sent to the message passing port of the XXXBlock block with id block_id.`

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
   repeated string block_id = 3;
 }
 
+// Command message sent to the flowgraph. This only works with the custom
```
anuraaga: `// This currently only works for the Starcoder XXX block which will forward the messages as is to any other block it is connected to.`

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
...
+    PMT key = 1;
+
+    // Value assigned to key
+    PMT value = 2;
+  }
+
+  repeated Entry entries = 1;
+}
+
+// Protobuf format of a PMT Vector
```
anuraaga: Protobuf isn't really a format, just a way of defining an API / protocol (serialized protobuf is a format). I think we can replace `Protobuf format of a` with 'A' in the comments.

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
...
+
+    C32Vector c32_value = 11;
+
+    C64Vector c64_value = 12;
+  }
+}
+
+// Protobuf format of a PMT Uint8 Vector
+message U8Vector {
+  bytes payload = 1;
```
anuraaga: I guess `value` throughout. Also, since we're using `int32` types for all the other ones, is there a reason to use `bytes` here?

reiinakano: Won't it be more efficient to use `bytes` instead of `uint32` here?

anuraaga: All the `int` types are variable encoded so it's the same for values [0, 127]. I think it's mainly whether users would find it surprising if the generated `go` code would be quite different (in Java, some types taking numbers as a parameter but just this one taking bytes would be very surprising).

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
...
+
+    UniformVector uniform_vector_value = 9;
+
+    Dict dict_value = 10;
+  }
+}
+
+// Protobuf format of a PMT pair. The `car` and `cdr` variable names are defined by
+// GNURadio.
+message Pair {
```
kenichi-fukushima: I think these new messages are worth their own package, e.g., `starcoder.pmt`. Otherwise, you should prefix them with `Pmt` because these terms like pair, tuple, dict and vector are too generic.

api/starcoder.proto
```diff
@@ -41,6 +52,163 @@ message RunFlowgraphRequest {
...
+
+    // Value assigned to key
+    PMT value = 2;
+  }
+
+  repeated Entry entries = 1;
+}
+
+// Protobuf format of a PMT Vector
+message Vector {
```
kenichi-fukushima: How is it different from `Tuple`?

api/starcoder.proto
kenichi-fukushima: Protobuf wire format varint. uint32 does not always takes 4 bytes when serialized
https://developers.google.com/protocol-buffers/docs/encoding

api/starcoder.proto
reiinakano: A tuple is immutable, and can only hold 10 PMTs (weird GNURadio behavior). A vector is mutable and has no size limit.

api/starcoder.proto
reiinakano: @anuraaga actually asked me to remove the `Pmt` prefix from these types. Wdyt?

api/starcoder.proto
```diff
@@ -41,6 +52,166 @@ message RunFlowgraphRequest {
...
+    F64Vector f64_value = 10;
+
+    C32Vector c32_value = 11;
+
+    C64Vector c64_value = 12;
+  }
+}
+
+// A PMT Uint8 Vector
+message U8Vector {
```
kenichi-fukushima: What about squashing these U8, U16, and U32 into one type with an enum for distinguishing between size variants? They all use 32 bit integer payload. They does not provide type-safety and the client code must perform unsafe cast to retrieve values anyway. And when doing the cast, it is not the compiler (unless you write a type-safe wrapper library) but the programmer who determines the right size to cast to. And there is no difference between determining the size from message type and doing it from the enum value.

api/starcoder.proto
kenichi-fukushima: I offered a suggestion with an argument for it. It is up to you as the code author whether to take that suggestion :)

anuraaga: The reason I don't see the need for `PMT` is that `PMT` is just a feature of how the structs are used by the `SendCommandRequest` - there's no need it has to be restricted to `PMT` now or in the future as the fact that these correspond to gnuradio type is more an implementation detail than an aspect of the API IMO.

reiinakano:
I squashed vectors with element size <=32 bits to one vector type. Not sure about the naming of the new messages. PTAL.

api/starcoder.proto
kenichi-fukushima: Maybe that's worth commenting? Are you going to the type information (distinction between these `Vector` and `Tuple`) to decide what PMT type to map to?

reiinakano: Yes, I'm going to the type info to decide which PMT type to map to. We could make another enum choosing either Vector or Tuple, though. But I'm not sure which is better for protobuf. Wdyt?

kenichi-fukushima: I would prefer one type with enum for consistency with `UVector`.

You can revisit the decision later after you write business logic code around this API and develop more insight about how it should look like.

PR #46, opened 2018-05-28
STELLAR-1173 Matplotlib oo api

reiinakano:
Uses matplotlib OO API to plot waterfall

### DIFF
--- a/gr-starcoder/python/plot_waterfall.py
+++ b/gr-starcoder/python/plot_waterfall.py
@@ -38,6 +38,9 @@
 import matplotlib
 matplotlib.use('agg')
 
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
+
 import io
 import numpy as np
 import matplotlib.pyplot as plt
@@ -61,7 +64,7 @@ def float_to_str(f):
     return '{:.1}'.format(f)
 
 
-def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
+def old_plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
 
     plt.figure(figsize=(12.8, 20.8), dpi=200)
     plt.imshow(arr, cmap=cm.nipy_spectral,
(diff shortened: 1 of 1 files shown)


gr-starcoder/python/plot_waterfall.py
```diff
@@ -63,11 +61,15 @@ def float_to_str(f):
 
 def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
 
-    plt.figure(figsize=(12.8, 20.8), dpi=200)
-    plt.imshow(arr, cmap=cm.nipy_spectral,
+    fig = Figure(figsize=(12.8, 20.8), dpi=200)
+    FigureCanvas(fig)
+
+    ax = fig.add_subplot(111)
```
anuraaga: What's 111? If it's a pixel location, is it possible to calculate it based on the width/height/dpi?

reiinakano: https://matplotlib.org/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.add_subplot

It means 1st position on a grid with 1 column and 1 row. It's not a pixel location, just means that we're not plotting more than one subplot.

gr-starcoder/python/plot_waterfall.py
```diff
@@ -63,11 +61,15 @@ def float_to_str(f):
 
 def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
 
-    plt.figure(figsize=(12.8, 20.8), dpi=200)
-    plt.imshow(arr, cmap=cm.nipy_spectral,
+    fig = Figure(figsize=(12.8, 20.8), dpi=200)
+    FigureCanvas(fig)
```
gavinreaney: Is this needed? It seems a bit strange to create this object then not do anything with it. Otherwise LGTM.

reiinakano: https://matplotlib.org/gallery/api/agg_oo_sgskip.html has a comment saying
> A canvas must be manually attached to the figure (pyplot would automatically do it).  This is done by instantiating the canvas with the figure as argument.

Bit weird I agree

PR #48, opened 2018-05-30
STELLAR-1193 Flip waterfall tick marks

reiinakano:
Follow up to previous PR. I needed to flip the y-axis tick marks as well.


gr-starcoder/python/plot_waterfall.py
```diff
@@ -101,8 +101,8 @@ def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
         ax.set_yticklabels(y_tick_labels, size='x-large')
     ax.grid(b='on', linestyle='dashed', linewidth=1,
             color='#000000', alpha=0.3)
-    minor_locator_y = IndexLocator(10 * rps, len(arr) % (10 * rps))
-    ax.get_yaxis().set_minor_locator(minor_locator_y)
+    ax.get_yaxis().set_minor_locator(IndexLocator(10 * rps, 0))
+    ax.get_yaxis().set_major_locator(IndexLocator(60 * rps, 0))
```
gavinreaney: Do you need to explicitly set the major locator now even although it wasn't set before?

reiinakano: Yeah. I actually think it wasnt properly aligned before either. It wasnt noticeable at all before though, something about having 0 at the bottom made it stand out a bit

reiinakano:
> When we changed the origin to be bottom left, did that cause the tick labels to flip their origin to top-right? I guess so and on that basis the change looks good.

No, the tick labels were top to bottom before, so it was actually aligned with the waterfall. When I flipped the image, the ticks didn't follow it. Now the issue should be fixed.

gavinreaney:
Are you certain? Previously Andrew sent me this link which has tick marks from bottom to top (well before your fix):
https://console.cloud.google.com/storage/browser/telemetry-data-dev/satellites/1/antennas/9/507904/WATERFALL?project=infostellar-cluster

That was what prompted me to ask for the data orientation to be checked because it seemed to not match.

reiinakano:
Oh, you're absolutely right.

> When we changed the origin to be bottom left, did that cause the tick labels to flip their origin to top-right? I guess so and on that basis the change looks good.

is correct then. My bad! :)

PR #51, opened 2018-06-15
STELLAR-1224 Add NOAA decoder and flowgraph

reiinakano:
Copied NOAA decoder from Satnogs.

NOAA flowgraph here:

screenshot from 2018-06-15 15-47-42


build.gradle
```diff
@@ -287,6 +287,7 @@ envs {
             condaPackage('subprocess32'),
             condaPackage('swig'),
             condaPackage('wget'),
+            condaPackage('libpng'),
```
anuraaga: Can you up the cache key version?

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,299 @@
...
+    }
+    // Write out the flipped image
+    if (d_filename_png != "") flipped.write(filename);
+    if (string_queue_ != NULL) {
+      std::ostringstream stream;
+      flipped.write_stream(stream);
+      string_queue_->push(stream.str());
+    }
+  }
+      // In case the flip option is not set
```
anuraaga: I think this comment is causing the formatter to get confused, can you move it inside the `else`?

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,297 @@
...
+    write_image(d_full_image, d_filename_png);
+  }
+  return true;
+}
+
+void noaa_apt_sink_impl::set_pixel(size_t x, size_t y, float sample) {
+  // Adjust dynamic range, using minimum and maximum values
+  sample = (sample - f_min_level) / (f_max_level - f_min_level) * 255;
+  // Set the pixel in the full image
+  d_full_image.set_pixel(x, y, sample);
```
anuraaga: Do you think we can migrate to use `libpng` directly? We don't use many functions so it might not be much more complicated. Adding `png++` will be annoying...

reiinakano: The function I need in Boost GIL (writing a PNG file to a stream) is in the develop branch. (and has been there for like 10 years...) https://github.com/boostorg/gil https://github.com/boostorg/gil/tree/master https://www.boost.org/doc/libs/develop/libs/gil/io/doc/html/index.html

We'll still probably need to work around the Conda limitation since Boost from there isn't the development version. Wdyt?

anuraaga: Not too sure, but when I look at the headers in `boost` in the conda prefix, I see `png_io` stuff. Does it not work?

```
ls /home-unencrypted/anuraag/.gradle/curiostack/python/bootstrap/miniconda2-4.5.4/include/boost/gil/extension/io
dynamic_io.hpp  jpeg_dynamic_io.hpp  jpeg_io_private.hpp  png_io.hpp          tiff_dynamic_io.hpp
io_error.hpp    jpeg_io.hpp          png_dynamic_io.hpp   png_io_private.hpp  tiff_io.hpp
```

reiinakano: No, those are the 10 year old ones :P

anuraaga: Ok so despite boost being the latest version, gil is very old. Will look into this a bit more. We're using conda since it allows the build to work cross-platform, very easily, and it'd be cool if we can get that to work here. Will look into seeing if there's a good C++ solution, alternatively we can just return a bitmap to the RPC client, or we can also do the image creation in python like for waterfalls.

anuraaga:
I think adding conda recipes would mean maintaining a package repository, which is annoying. We can add pybombs recipes though which will just build the library automatically during setup.

gr-starcoder/lib/noaa_apt_sink_impl.h
```diff
@@ -0,0 +1,122 @@
...
+  // Writes a single image to disk, also takes care of flipping
+  void write_image(std::string filename);
+
+  // Factor exponential smoothing average,
+  // which is used for sync pattern detection
+  const float f_average_alpha;
+  static const bool synca_seq[];
+  static const bool syncb_seq[];
+
+  std::string d_filename_png;
```
kenichi-fukushima: what is the `d_` prefix?

gr-starcoder/lib/noaa_apt_sink_impl.h
```diff
@@ -0,0 +1,122 @@
...
+
+  std::string d_full_filename;
+  std::string d_left_filename;
+  std::string d_right_filename;
+
+  size_t d_current_x;
+  size_t d_current_y;
+  size_t d_num_images;
+
+  float f_max_level;
```
kenichi-fukushima: what's the `f_` prefix?

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,302 @@
...
+  d_current_x = new_x;
+}
+
+noaa_apt_sync_marker noaa_apt_sink_impl::is_marker(size_t pos,
+                                                   const float *samples) {
+  // Initialize counters for 'hacky' correlation
+  size_t count_a = 0;
+  size_t count_b = 0;
+
+  for (size_t i = 0; i < 40; i++) {
```
kenichi-fukushima: Define a constant with self-explanatory name for this number.

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,302 @@
...
+    if ((sample > 0 && synca_seq[i]) || (sample < 0 && !syncb_seq[i])) {
+      count_a += 1;
+    }
+    if ((sample > 0 && syncb_seq[i]) || (sample < 0 && !syncb_seq[i])) {
+      count_b += 1;
+    }
+  }
+
+  // Prefer sync pattern a as it is detected more reliable
+  if (count_a > 35) {
```
kenichi-fukushima: define a constatnt

gr-starcoder/lib/noaa_apt_sink_impl.h
```diff
@@ -0,0 +1,122 @@
...
+  bool stop();
+  void register_starcoder_queue(uint64_t ptr);
+
+ private:
+
+  /*
+   * Checks if the history portion of the input contains a sync marker.
+   * Matches the 40 samples before pos against the patterns.
+   */
+  noaa_apt_sync_marker is_marker(size_t pos, const float *samples);
```
kenichi-fukushima: Seems this function can be `const`

anuraaga:
I think so far we haven't been changing style of forked open-source code, outside of the whitespace formatting so it's easier to compare with upstream

kenichi-fukushima:
Ah, I didn't notice it was a fork. Then, you should have ideally checked in the original file without modification in one PR and made customization in another so that diff from the original is obvious.

reiinakano:
Sorry, the majority of this code was copied from Satnogs' APT decoder. The changes I made were mostly in the `write_image` function. I will address your comments there.

gr-starcoder/include/starcoder/noaa_apt_sink.h
```diff
@@ -0,0 +1,87 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this software; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#ifndef INCLUDED_STARCODER_NOAA_APT_SINK_H
+#define INCLUDED_STARCODER_NOAA_APT_SINK_H
+
+#include <starcoder/api.h>
```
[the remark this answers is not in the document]
reiinakano: This is autogenerated by GNURadio and is present in all blocks. The `STARCODER_API` constant used below is defined in this header, but admittedly, I'm not 100% sure what it does.

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,302 @@
...
+ * along with this software; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#define png_infopp_NULL (png_infopp) NULL
```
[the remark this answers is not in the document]
reiinakano: This is needed for https://github.com/ignf/gilviewer/issues/8

gr-starcoder/lib/noaa_apt_sink_impl.cc
```diff
@@ -0,0 +1,302 @@
...
+
+    boost::gil::detail::png_writer w(temp.native().c_str());
+
+    if (!d_flip)
+      w.apply(const_view(image_received_));
+    else
+      w.apply(flipped_up_down_view(view(image_received_)));
+
+    std::ifstream t(temp.native());
+    std::stringstream buffer;
```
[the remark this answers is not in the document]
reiinakano: I tried this but it seems to break the generated PNG during runtime.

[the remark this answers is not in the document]
kenichi-fukushima: That finding should be worth commenting. As a rule of thumb, please comment any of your decisions that the code itself can't communicate to future maintainers.

PR #52, opened 2018-06-19
STELLAR-1231 Add Prometheus metrics export

reiinakano:
Right now only exports the number of flowgraphs currently running.

### DIFF
--- a/gr-starcoder/python/plot_waterfall.py
+++ b/gr-starcoder/python/plot_waterfall.py
@@ -36,6 +36,7 @@
 
 import io
 import numpy as np
+import gc
 
 from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
 from matplotlib.figure import Figure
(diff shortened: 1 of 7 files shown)


gr-starcoder/python/plot_waterfall.py
```diff
@@ -60,6 +61,7 @@ def float_to_str(f):
 
 
 def plot_waterfall(arr, samp_rate, center_freq, rps, fft, filename):
+    gc.collect()
```
anuraaga: Remove?

reiinakano: I've found this helps reduce memory usage, at least on local. I don't think this affects latency too much, considering that the waterfall plot already takes a while. Since it's just two lines, I put it here. I could open a separate PR for this?

anuraaga: Do we think the memory usage is from creating the waterfall? In that case, how about doing it at the end of this function?

reiinakano: Yeah, it always expands when the waterfall script starts to run. I think doing `collect` at the end of the function won't free the memory that we allocated in that particular function, since they'd still be in scope. By doing `gc.collect` at the start, we can force clean up resources left by the previous call to the waterfall plot.

I could *maybe* call `gc.collect` from C++ after the waterfall plot goes out of scope?

anuraaga: Ah that makes sense - anyways let's leave it like this then

PR #54, opened 2018-06-22
STELLAR-1237 Export GNURadio per-block performance counters to Prometheus

reiinakano:
Each GNURadio block provides performance counters that can be polled while the flowgraph is running. https://wiki.gnuradio.org/index.php/PerformanceCounters

Since we can't use the CQueue this time short of messing with core GNURadio, we'll have to go through Python to retrieve them, but I think this is fine since we'll only run it once every ~15 seconds.


monitoring/metrics.go
```diff
@@ -46,3 +58,23 @@ func (m *Metrics) init() {
 func getMetricsName(name string) string {
 	return metricsPrefix + name
 }
+
+// Registers a new gauge. If gauge already exists, returns that.
+func GetPerfCtrGauge(flowgraphName string, blockName string, counterName string) (prometheus.Gauge, error) {
+	gauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: getMetricsName(flowgraphName + "_" + blockName + "_" + counterName),
```
anuraaga: `blockName` and `counterName` can be labels instead of using string manipulation like this

server/starcoder.go
```diff
@@ -415,10 +496,11 @@ func (s *Starcoder) compileGrc(path string) (*moduleAndClassNames, error) {
 	return retVal, nil
 }
 
-func (s *Starcoder) startFlowGraph(modAndImport *moduleAndClassNames, request *pb.StartFlowgraphRequest) (*python.PyObject, map[string]*cqueue.CStringQueue, map[string]*cqueue.CStringQueue, error) {
+func (s *Starcoder) startFlowGraph(modAndImport *moduleAndClassNames, request *pb.StartFlowgraphRequest) (*python.PyObject, map[string]*cqueue.CStringQueue, map[string]*cqueue.CStringQueue, map[string]*python.PyObject, error) {
```
anuraaga: It might make sense to create a type to contain all these return values. It seems like a lot of related state being passed around in multiple variables.

server/starcoder.go
```diff
@@ -238,6 +247,69 @@ func newStreamHandler(sc *Starcoder, stream pb.Starcoder_RunFlowgraphServer, fgI
 		}()
 	}
 
+	// Processing loop for performance counters
```
anuraaga: Extract a method to prevent this getting too long (probably good to extract the others too).

monitoring/metrics.go
reiinakano: Was using `constLabels`, which is a bit different than regular labels. Oops :)

server/starcoder.go
```diff
@@ -35,6 +35,8 @@ import (
 	"go.uber.org/zap"
 	"github.com/gogo/protobuf/proto"
 	"github.com/infostellarinc/starcoder/monitoring"
+	"time"
+	"github.com/prometheus/client_golang/prometheus"
```
anuraaga: Alphabetical  / gofmt

PR #60, opened 2018-07-03
STELLAR-1248 Meteor flowgraph

reiinakano:
The actual meteor flowgraph

screenshot from 2018-07-03 20-10-48

.. and some missed edge case handling in C++

### DIFF
--- a/sampleclient/main.go
+++ b/sampleclient/main.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"log"
 	"time"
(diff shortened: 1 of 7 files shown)


sampleclient/main.go
```diff
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"log"
 	"time"
+	"fmt"
```
anuraaga: alphabetical

PR #63, opened 2018-07-26
STELLAR-1371 Strip ax25 header from decoder

reiinakano:
So we can start testing the simulation UDP software (?) from BA.

Adds a new parameter to the AX.25 Decoder block "Strip Headers"


gr-starcoder/lib/ax25_decoder_bm_impl.cc
```diff
@@ -342,9 +345,14 @@ void ax25_decoder_bm_impl::enter_frame_end() {
              d_frame_buffer[d_received_bytes - 2];
 
   if (fcs == recv_fcs) {
-    message_port_pub(
-        pmt::mp("pdu"),
-        pmt::make_blob(d_frame_buffer, d_received_bytes - sizeof(uint16_t)));
+    int padding = 0;
```
anuraaga: Usually called `offset`

PR #64, opened 2018-07-27
Use pybombs boost instead of conda.

anuraaga:
Too lazy to try build on my machine so will see what PR build does.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,11 +27,11 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1.3-dependencies
+          - v1.4-dependencies
 
       - run: ./gradlew --no-daemon install
 
       - save_cache:
           paths:
             - ~/.gradle
-          key: v1.3-dependencies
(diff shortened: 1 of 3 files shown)


reiinakano:
Any ideas what's failing here? I've been trying to reproduce it but I'm getting a different PyBOMBS `AssertionError`... Is there any special benefit we're getting from using PyBOMBS? Do you think `make install`ing everything might make things simpler?

anuraaga:
We use PyBombs because it's the official build tool of GnuRadio and allows for a flexible, cross-platform build.

We don't use `make install` since it requires developers to modify their systems to build, which makes builds unreproducible and undebuggable, and therefore we can't provide any support at all to developers.

I found that an update to a pybombs dependency, `ruamel.yaml`, broke it - the fix for that is in master already but hasn't been released. To workaround, fixing `ruamel.yaml` to a lower version fixes that it seems. I also saw both `uhd` and `gnuradio` have fixes for building with latest boost version (it was actually conda being newer than supported not older, oops) so checking now if they build on my machine and will send the fix.

PR #65, opened 2018-08-14
Update gnuradio / uhd to latest versions for boost compatibility.

anuraaga:
Also fixes pybombs build by fixing version of yaml parser. Confirmed on my machine gnuradio builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,8 +29,9 @@ jobs:
           keys:
           - v1.4-dependencies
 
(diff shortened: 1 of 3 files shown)


reiinakano:
Looks like it got stuck at 100% then received the killed signal. According to https://discuss.circleci.com/t/kill-signal-problem/18665/6, the build might be using too much memory.

anuraaga:
Not sure why this is freezing in CI, maybe a PyBombs issue... Going to merge and see what happens on quay

PR #70, opened 2018-08-27
STELLAR-1528 Fix noaa corrupt png

reiinakano:


danada:
Wicked! Is it possible to produce a sample image I can test just to make sure this is fixed in safari too?

reiinakano:
noaa_apt_rec

@danada could you check this one?

PR #74, opened 2018-08-30
STELLAR-1570 run uhd images downloader

reiinakano:
UHD needs to download some additional images after installation to properly operate the USRP. We did this by directly running
`/root/.gradle/curiostack/python/bootstrap/miniconda2-gnuradio/lib/uhd/utils/uhd_images_downloader.py`
in the image.

Dunno if it'll work, let's see what Circle does.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1.4-dependencies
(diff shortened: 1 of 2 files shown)


reiinakano:
"Received kill signal" after just 40 minutes.

reiinakano:
@anuraaga if the PR looks okay with you, I'll merge and see what happens on Quay. It looks like Circle won't cooperate

PR #75, opened 2018-09-07
Migrate to go modules.

anuraaga:


anuraaga:
Since this configuration is our only build config, nothing else to change!

PR #76, opened 2018-09-10
Add configuration for suppressing particular command logs

reiinakano:
Logging that a certain block in Starcoder received a command is very useful, but for blocks like Doppler shift which receive a command hundreds of times a second, very spammy. This PR adds a configuration option to silence logs for a particular command block.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -54,6 +54,7 @@ type serveCmdConfiguration struct {
 	FlowgraphDir              string
 	ExporterAddress           string
 	PerfCtrCollectionInterval time.Duration
+	SilencedCommandBlocks     []string
 }
 
 var serveCmdConfig = serveCmdConfiguration{}
@@ -71,6 +72,7 @@ var serveCmd = &cobra.Command{
 		log := l.Sugar()
 		defer log.Sync()
 
+		log.Infof("Using configuration: %+v", serveCmdConfig)
 		log.Infof("serve called, using bind address %v", serveCmdConfig.BindAddress)
 
 		// Set up metrics endpoint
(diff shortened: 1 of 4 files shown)


cmd/serve.go
```diff
@@ -167,10 +169,11 @@ func init() {
 
 	// Cobra supports local flags which will only run when this command
 	// is called directly, e.g.:
-	serveCmd.Flags().StringVar(&serveCmdConfig.BindAddress, "bind-address", defaultBindAddress, "Address to bind to")
-	serveCmd.Flags().StringVar(&serveCmdConfig.FlowgraphDir, "flowgraph-dir", "", "Directory containing GNURadio flowgraphs to serve. If blank, defaults to built-in Starcoder flowgraphs.")
-	serveCmd.Flags().StringVar(&serveCmdConfig.ExporterAddress, "exporter-address", defaultExporterAddress, "Address where exported Prometheus metrics will be served")
-	serveCmd.Flags().DurationVar(&serveCmdConfig.PerfCtrCollectionInterval, "perf-ctr-interval", defaultPerfCtrCollectionInterval, "Time interval for exporting GNURadio performance metrics to Prometheus. If set to 0, this will be disabled. Default 15s.")
+	serveCmd.Flags().StringVar(&serveCmdConfig.BindAddress, "bind_address", defaultBindAddress, "Address to bind to")
```
anuraaga: I've never seen a golang program use underscores in flag names. Does viper not automatically convert dashes to underscores?

anuraaga: Found that it's probably conventional to use SetEnvKeyReplacer to replace `-` to `_`

reiinakano: Oops didn't notice, thanks, will followup

PR #78, opened 2018-09-10
Add functionality so Starcoder can send back arbitrary PMTs, not just blobs

reiinakano:
Adds functionality so Starcoder can send back arbitrary PMTs, not just data bytes. Instead of sending a serialized PMT, we convert it to the appropriate serialized Protobuf format and send it up the string queue. 

We also introduce the Blob PMT. On the GNURadio side, this is equivalent to a uint8_vector, but since our protobuf maps the uint8_vector to a repeated uint32 field, it introduces a lot of overhead and necessitates weird conversion code when working with it from the gRPC side (e.g. converting `[]uint32` to `[]byte` in Go).


server/starcoder.go
```diff
@@ -239,37 +239,55 @@ func (sh *streamHandler) observableQueueLoop(blockName string, q *cqueue.CString
...
+				if err != nil {
+					sh.log.Errorw("Failed to unmarshal protobuf", "block", blockName)
 					continue
 				}
-				if err := sh.stream.Send(&pb.RunFlowgraphResponse{
+				request := &pb.RunFlowgraphResponse{
 					BlockId: blockName,
-					Payload: bytes,
-				}); err != nil {
+					Pmt:     message,
```
vaelen: It looks like this code is duplicated above.  Could this be made more DRY?

cqueue/string_queue.h
```diff
@@ -42,6 +42,7 @@ class string_queue {
     std::string blocking_pop();
     unsigned long get_ptr() const;
     void close();
+    bool closed();
```
kenichi-fukushima: This is semantically `const`. Not sure if the compiler allows it to be `const` if it acquires and releases a mutex though.

gr-starcoder/lib/pmt_to_proto.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef INCLUDED_PMT_TO_PROTO_H
+#define INCLUDED_PMT_TO_PROTO_H
+
+#include <pmt/pmt.h>
+#include "starcoder.pb.h"
+
+void convert_pmt_to_proto(const pmt::pmt_t &pmt_msg,
+                          starcoder::BlockMessage *proto_msg);
+void convert_proto_uniform_vector(const pmt::pmt_t &pmt_msg,
```
kenichi-fukushima: It seems like this function is only used by `convert_pmt_to_proto` and does not need to be exposes in a header file.

gr-starcoder/lib/pmt_to_proto.h
```diff
@@ -0,0 +1,32 @@
...
+ * Boston, MA 02110-1301, USA.
+ */
+
+#ifndef INCLUDED_PMT_TO_PROTO_H
+#define INCLUDED_PMT_TO_PROTO_H
+
+#include <pmt/pmt.h>
+#include "starcoder.pb.h"
+
+void convert_pmt_to_proto(const pmt::pmt_t &pmt_msg,
```
kenichi-fukushima: Feel free to pass `BlockMessage` back as a return value. I believe the compiler is smart enough to avoid copying an object when returning it (and actually this behavior is required by the language specification.)

gr-starcoder/lib/pmt_to_proto.cc
```diff
@@ -0,0 +1,171 @@
...
+    const std::vector<double> vector_elements =
+        pmt::f64vector_elements(pmt_msg);
+    *f64_vector->mutable_value() = { vector_elements.begin(),
+                                     vector_elements.end() };
+  } else if (pmt::is_c32vector(pmt_msg)) {
+    starcoder::C32Vector *c32_vector = uni_vector->mutable_c32_value();
+    const std::vector<std::complex<float>> vector_elements =
+        pmt::c32vector_elements(pmt_msg);
+    std::transform(vector_elements.begin(), vector_elements.end(),
+                   c32_vector->mutable_value()->begin(),
```
kenichi-fukushima: Don't you need to use a back inserter? Is it guaranteed that `c32_vector` has enough buffer to store all transformed elements?
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.repeated_field

gr-starcoder/lib/pmt_to_proto.cc
```diff
@@ -0,0 +1,171 @@
...
+      new_val.set_real_value(c.real());
+      new_val.set_imaginary_value(c.imag());
+      return new_val;
+    });
+  } else if (pmt::is_c64vector(pmt_msg)) {
+    starcoder::C64Vector *c64_vector = uni_vector->mutable_c64_value();
+    const std::vector<std::complex<double>> vector_elements =
+        pmt::c64vector_elements(pmt_msg);
+    std::transform(vector_elements.begin(), vector_elements.end(),
+                   c64_vector->mutable_value()->begin(),
```
kenichi-fukushima: same as above

gr-starcoder/lib/pmt_to_proto.h
reiinakano: This is what I did first. The problem is the C++ generated protobuf has no way to set embedded message fields other than through getting a pointer to the embedded message field with the `mutable_xxx()` method and directly modifying that pointer. For the recursion to work, not sure if there's another way than having the function take a pointer.

cqueue/string_queue.h
reiinakano: Nope :(

gr-starcoder/lib/pmt_to_proto.h
kenichi-fukushima: It seems you can call `Swap()` on a mutable message. I haven't verified this API on my side and leave the decision up to you.
https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#message

```
AMessageType parentMessage;

...
AMessageType submessage = convert_pmt_to_proto(...);
parentMessage.get_mutable_field()->Swap(&submessage);

....

return parentMessage;
```

gr-starcoder/lib/pmt_to_proto.cc
reiinakano: Okay, I modified it. Could you check if I did it correctly? Sorry, I don't fully understand STL containers yet...

gr-starcoder/lib/pmt_to_proto.h
reiinakano: I guess this would work. Is there a specific reason to prefer returning the `BlockMessage` though?

anuraaga: Just for reference, found these about the protobuf move semantics. Not sure if copy when reallocating affects us 

https://github.com/protocolbuffers/protobuf/issues/2791
https://github.com/protocolbuffers/protobuf/issues/3630

gr-starcoder/lib/pmt_to_proto.cc
```diff
@@ -0,0 +1,175 @@
...
+#include "pmt_to_proto.h"
+
+void convert_proto_uniform_vector(const pmt::pmt_t &pmt_msg,
+                                  starcoder::UniformVector *uni_vector) {
+  if (pmt::is_u8vector(pmt_msg)) {
+    starcoder::UVector *u_vector = uni_vector->mutable_u_value();
+    u_vector->set_size(starcoder::IntSize::Size8);
+    const std::vector<uint8_t> vector_elements =
+        pmt::u8vector_elements(pmt_msg);
+    *u_vector->mutable_value() = { vector_elements.begin(),
```
kenichi-fukushima: My C++ knowledge hasn't been update for C++11 and I really don't know how initializer expressions work... Can you explain how this assignment works? My reasoning is the right hand side calls `RepeatedPtrField::RepeatedPtrField(Iter begin, const Iter & end)` and then the whole content is copied for the another time, into the left hand side, which is redundant. You used `transform` and an inserter below. Why this piece of code follows a different pattern?

gr-starcoder/lib/pmt_to_proto.h
kenichi-fukushima: FYI. No explicit move is needed in this case. Its optimization for returning a local variable from a function.
https://gist.github.com/kenichi-fukushima/197238e0cd23b82666a27abfa5fbd540

gr-starcoder/lib/pmt_to_proto.cc
reiinakano: I changed it to use `std::copy`. Wdyt?

kenichi-fukushima: Looks good :+1:

gr-starcoder/lib/pmt_to_proto.h
reiinakano: @kenichi-fukushima  Converted this to your suggested function signature. It does seem more intuitive to use.

PR #79, opened 2018-09-10
install git on builder

reiinakano:
Dunno if this is the correct solution, but I noticed the Circle Docker image installs git while tools/builder/Dockerfile doesn't.


reiinakano:
ping

anuraaga:
Almost done with rewriting the build tasks for this which should solve it will send it soon.

reiinakano:
Ahh got it

PR #85, opened 2018-09-13
Fix run.sh

reiinakano:
looks like it adds another folder with the conda version


reiinakano:
Actually, this is still wrong. I can't find /root/.gradle/curiostack/gnuradio/4.4.10/bin/starcoder at all. There's no binary called `starcoder` in the whole container.

PR #87, opened 2018-09-14
Add gr-satellites

reiinakano:
Currently, we only need one block from this but I imagine we'll need more in the future. Not sure if it's better to do this or to copy and paste what we need like before. Wdyt?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -271,7 +271,9 @@ task installGrSatellites(type: org.curioswitch.gradle.conda.tasks.CondaExecTask)
 
     condaName 'gnuradio'
 
-    command "pybombs -y install gr-satellites"
+    command """pybombs -y recipes add-defaults && \\
+        pybombs -y install gr-satellites
+    """
 }
 
 task installGrStarcoder(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
--- a/build.gradle
+++ b/build.gradle
@@ -271,7 +271,10 @@ task installGrStarcoder(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) 
 
     condaName 'gnuradio'
 
-    command "pybombs -vv -y install gr-starcoder"
+    command """pybombs -vv config default_prefix gnuradio && \\
+        pybombs -vv -y recipes add gr-starcoder ${project.file('gr-recipes')} && \\
+        pybombs -vv -y install gr-starcoder
+"""
 
     execCustomizer {exec->
         exec.environment 'PKG_CONFIG_PATH', "${gnuradioRoot}/lib/pkgconfig"
--- a/build.gradle
+++ b/build.gradle
@@ -266,7 +266,7 @@ task setupPrefix(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
     }
 }
 
-task installGrSatellites(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
+task installGrStarcoder(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
     dependsOn setupPrefix
 
     condaName 'gnuradio'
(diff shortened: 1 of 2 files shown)


anuraaga:
It's better to use pybombs than copy-paste (if we can use pybombs for others we should switch to that).

The best pattern would be to finally add a recipe for starcoder in `gr-recipes`, `gr-starcoder.lwr`, which depends on `gnuradio-nogui` and `gr-satellites` and runs `cmake` (should happen automatically by extending the `cmake` recipe like most block recipes out there). Then we replace the command in `installGrCStarcoder` from a cmake invocation to `pybombs install gr-starcoder`. Do you want to try that out?

reiinakano:
Yeah that sounds better. I guess we would also need our own recipe of gr-satellites since it depends on gnuradio when we want to make it depend on gnuradio-nogui.

anuraaga:
I think we can set `packages.gnuradio.forceinstalled: true` in our recipe to prevent anyone we use from pulling in the monster.

reiinakano:
This'll probably fail because gr-starcoder is in a subdirectory of starcoder repo. I tried skimming through the code and examples and I didn't see a way to specify a subdirectory. Do you know a good way of solving this?

anuraaga:
Hmm - I don't think it should matter since we set that actual subdirectory as a recipe path
`pybombs -y recipes add gr-starcoder ${project.file('gr-recipes')}`

Does that cause issues?

anuraaga:
By the way, if you mean how would other users do it, I was thinking we'll probably need to make a separate `gr-recipes` repository once we're ready. First let's get it working in this build and then we'll move it.

reiinakano:
I don't mean adding the recipe, I mean in the recipe gr-starcoder.lwr itself, I'm setting `source: git+https://github.com/infostellarinc/starcoder`.

anuraaga:
Also actually when looking though these files, I realized that we use the normal `gnuradio` recipe, just reconfigure it for our prefix. Users that just use normal `gnuradio` don't need to worry about it so the recipe should work both with our special `nogui` prefix or just a standard gnuradio install.

build.gradle
```diff
@@ -271,19 +271,15 @@ task installGrStarcoder(type: org.curioswitch.gradle.conda.tasks.CondaExecTask)
 
     condaName 'gnuradio'
 
-    command """cmake ../../gr-starcoder -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=${gnuradioRoot} -Wno-dev && \\
-        make &&
-        make install
+    command """pybombs -vv config default_prefix gnuradio && \\
+        pybombs -vv -y recipes add-defaults && \\
+        pybombs -vv -y recipes add gr-starcoder ${project.file('gr-recipes')} && \\
+        pybombs -vv -y install gr-satellites
```
anuraaga: The prefix is already set up correctly so I don't think you need to call anything but `install gr-satellites` here.

anuraaga: Oh - I mean `gr-starcoder`

gr-recipes/gr-starcoder.lwr
```diff
@@ -0,0 +1,47 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with PyBOMBS; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+category: common
+depends:
+- gnuradio-nogui
```
anuraaga: `gnuradio-nogui` is actually a prefix, it doesn't work here and isn't really needed I guess.

gr-recipes/gr-starcoder.lwr
```diff
@@ -0,0 +1,47 @@
...
+- gr-satellites
+description: Recipe for installing the Starcoder OOT
+config:
```
anuraaga: Remove this, the prefix is already set up correctly with gnuradio-nogui

build.gradle
reiinakano: So I actually get this warning without configuring the default prefix first. 

PyBOMBS.ConfigManager.PrefixInfo - DEBUG - Cannot establish a prefix directory. This may cause issues down the line.

I still get an error with it but a different one right now..

gr-recipes/gr-starcoder.lwr
reiinakano: Do I remove the whole config? I still want to set pyconstruct to forceinstalled: true if possible

build.gradle
reiinakano: Oh I didn't npotice my typo, nvm

reiinakano:
So it really doesn't seem like the prefix from setupPrefix is getting recognized. If I try only `pybombs install gr-starcoder` it complains it can't find the recipe. Then after adding the line `pybombs -vv -y recipes add gr-starcoder ${project.file('gr-recipes')} &&` it complains it can't find gr-satellites. Then after adding `pybombs -vv -y recipes add-defaults &&` it doesn't seem to honor the forceinstalled packages anymore, which should have been setup by setupPrefix.

PR #88, opened 2018-09-14
Add a recipe for pybombs user to install gr-starcoder.

anuraaga:
Just needed to set `PYBOMBS_PREFIX`.

Also created a separate gradle task for now since installing with this command currently can only build our merged code, not unmerged code which we need for PR builds. I think we can override to use a local directory in `gnuradio-nogui.lwr` but haven't looked into it yet.

### DIFF
--- a/gr-recipes/gr-starcoder.lwr
+++ b/gr-recipes/gr-starcoder.lwr
@@ -1,0 +1,27 @@
+#
+# Starcoder - a server to read/write data from/to the stars, written in Go.
+# Copyright (C) 2018 InfoStellar, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


gr-recipes/gr-starcoder.lwr
```diff
@@ -0,0 +1,27 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+category: common
```
reiinakano: Whoops I forgot to check. This doesn't install gr-satellites does it?

anuraaga: Right - I just wanted to create a working recipe for now and we can add stuff to it. I guess we shoudl go ahead and add `gr-satellites` to both this recipe and `gnuradio-nogui`. We can think of this recipe as our public installation script while gnuradio-nogui is our custom one for creating a trim docker image.

PR #90, opened 2018-09-14
add gr-satellites

reiinakano:
I guess we'd also need to update cache for pyconstruct to be forceinstalled?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -195,7 +195,7 @@ task setupPrefix(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
     command """pybombs auto-config && \\
         pybombs -y recipes add gr-starcoder ${project.file('gr-recipes')} && \\
         pybombs -y recipes add-defaults && \\
-        pybombs -y -vv prefix init ${gnuradioRoot} -a gnuradio -R gnuradio-nogui && \\
+        pybombs -y -v prefix init ${gnuradioRoot} -a gnuradio -R gnuradio-nogui && \\
         ${gnuradioRoot}/lib/uhd/utils/uhd_images_downloader.py
     """
 


gr-recipes/gnuradio-nogui.lwr
```diff
@@ -89,6 +90,8 @@ config:
       forceinstalled: true
     numpy:
       forceinstalled: true
+    pyconstruct:
```
anuraaga: Do you need to add this to `pythonPackages`?

build.gradle
```diff
@@ -76,6 +76,7 @@ conda {
         ]
         pythonPackages = [
                 'cheetah',
+                'construct',
```
anuraaga: `pyconstruct` and `construct` are different I think

reiinakano: Yeah in gr-etcetera the recipe was named pyconstruct even if they were installing construct. We want construct here

anuraaga: Wow - ok

reiinakano:
I modified the task to actually install from recipe. Had to update cache since it couldn't find gr-starcoder recipe. Does that look right?

build.gradle
```diff
@@ -306,7 +307,7 @@ task installStarcoder(type: Copy) {
 }
 
 task install {
-    dependsOn installGrStarcoder, installStarcoder
+    dependsOn installGrStarcoderFromRecipe, installStarcoder
```
anuraaga: We shouldn't have to do this since we have `gr-satellites` in `gnuradio-nogui.lwr`. If we do this it's not clear which starcoder gets installed, the upstream one or the one in this directory.

reiinakano:
So libfec failed to install from pybombs. Same error I was debugging last night https://github.com/gnuradio/pybombs/issues/523

Should I modify `elevate_pre_args` in gnuradio-nogui to fix it?

anuraaga:
sudo doesn't preserve the environment so nothing will be found. But anyways, sudo isn't ever a correct workaround for a build issue like this. Sent #91

build.gradle
```diff
@@ -192,9 +193,9 @@ task setupPrefix(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
     condaName 'gnuradio'
 
     command """pybombs auto-config && \\
-        pybombs -y recipes add-defaults && \\
         pybombs -y recipes add gr-starcoder ${project.file('gr-recipes')} && \\
-        pybombs -y -v prefix init ${gnuradioRoot} -a gnuradio -R gnuradio-nogui && \\
+        pybombs -y recipes add-defaults && \\
+        pybombs -y -vv prefix init ${gnuradioRoot} -a gnuradio -R gnuradio-nogui && \\
```
anuraaga: Make sure to remove verbose before merging

PR #91, opened 2018-09-14
Add custom libfec recipe that uses latest source and cmake build supp…

anuraaga:
…ort.

We should probably send a PR to gr-etcetera at some point.


reiinakano:
Got it, incorporated

PR #95, opened 2018-09-26
STELLAR-1671 AAUSAT4 decoder

reiinakano:
This PR adds an AAUSAT-4 decoder. The flowgraph is shown below:

screenshot from 2018-09-26 13-53-57

The flowgraph consists of 6 stages:
* Initial decimation and Doppler correction from the radio
* FSK demodulation. AAUSAT4 uses 2.4kbaud
* Sync preamble detection
* CCSDS Viterbi decoding
* CCSDS Additive scrambling
* CCSDS Reed Solomon decoding

The fully error corrected packet is a CSP (https://en.wikipedia.org/wiki/Cubesat_Space_Protocol) packet with an extra 2 bytes at the beginning containing the packet length and 2 bytes at the end containing HMAC. 

The details of the error correction used + byte format is defined in http://www.space.aau.dk/aausat4/index.php?n=Main.Ham

Some more useful info about decoding AAUSAT4: 
* https://destevez.net/2016/04/first-signals-from-aausat-4/ 
* https://destevez.net/2016/05/a-bunch-of-data-from-aausat-4/

We also introduce a new block pdu_trim_uvector which simply trims out the first and/or last N bytes of the packet. We need this because of the unique format of the packet, which includes extra header information between certain steps of the forward error correction. It's written in Python since it's not a particularly hot path. We can port it to C++ if we find it necessary in the future.


vaelen:
Please wait for @kenichi-fukushima 's review as well before merging.

PR #102, opened 2018-11-13
IQ Only Receiver hierarchical block

reiinakano:
Main change here is addition of hierarchical block `iq_only_receiver`.

`iq_only_receiver` wraps `radio_source` but includes our logic for accepting Doppler shift information, shifting away the DC bias, and resampling (usually to 112.5k), taking into account the differences between the USRP and AR2300 (non-variable sampling rate, hardware frequency not controlled by GNU Radio, etc)

FSK flowgraph
fsk_9600 grc

Custom bandwidth
custom_bandwidth grc

AAUSAT
aausat4 grc


gr-starcoder/python/iq_only_receiver.py
```diff
@@ -0,0 +1,95 @@
...
+
+from gnuradio import gr
+from gnuradio import blocks
+from gnuradio import analog
+from gnuradio import filter
+from radio_source import radio_source
+
+class iq_only_receiver(gr.hier_block2):
+    """
+    docstring for block iq_only_receiver
```
kenichi-fukushima: Please provide a meaningful docstring.

PR #104, opened 2018-11-15
fix failing waterfall test by updating test image

reiinakano:
also updated the test code to tolerate some level of difference in the generated images


reiinakano:
The test is taking too long because `No cache is found for key: v2.1-dependencies`.

Did Circle just delete our cache randomly?

anuraaga:
Maybe :(

reiinakano:
@anuraaga this shows up in Circle

"Your current usage represents 198% of your infostellarinc Linux plan's limit. Please upgrade in order to ensure no disruption in building."

PR #109, opened 2018-11-16
Golay decoder block

kenichi-fukushima:
TODO
- [x] Actually call decode_golay24 in the block.
- [x] Rewrite the test in Python. Python tests produce much better information about test result.
- [x] Finish the block xml file.
- [x] Handle an error of too many errors to be corrected.


gr-starcoder/lib/golay24.c
```diff
@@ -0,0 +1,119 @@
+/*
```
kenichi-fukushima: Golay implementation is taken from gr-satellites.
https://github.com/daniestevez/gr-satellites/blob/master/lib/golay24.h

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+namespace starcoder {
+
+golay_decoder::sptr golay_decoder::make(int offset, int num_units) {
+  return gnuradio::get_initial_sptr(new golay_decoder_impl(offset, num_units));
+}
+
+golay_decoder_impl::golay_decoder_impl(int offset, int num_units)
+    : gr::block("golay_decoder", gr::io_signature::make(0, 0, 0),
+                gr::io_signature::make(0, 0, 0)),
+      _offset(offset),
```
reiinakano: I think we use `offset_` for member variables

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+      _num_units(num_units) {
+  message_port_register_in(pmt::mp("in"));
+  message_port_register_out(pmt::mp("out"));
+  set_msg_handler(pmt::mp("in"),
+                  boost::bind(&golay_decoder_impl::msg_handler, this, _1));
+}
+
+golay_decoder_impl::~golay_decoder_impl() {}
+
+void golay_decoder_impl::forecast(int noutput_items,
```
reiinakano: No need to implement `forecast` or `general_work` I think.

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+int golay_decoder_impl::general_work(int noutput_items,
+                                     gr_vector_int &ninput_items,
+                                     gr_vector_const_void_star &input_items,
+                                     gr_vector_void_star &output_items) {
+  return 0;
+}
+
+void golay_decoder_impl::msg_handler(pmt::pmt_t pmt_msg) {
+  size_t in_size(0);
+  const uint8_t *in = pmt::u8vector_elements(pmt::cdr(pmt_msg), in_size);
```
reiinakano: Just confirming, does the size of the `u8_vector` get assigned to `in_size`? Never used that function signature before.

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,80 @@
...
+
+  std::vector<uint8_t> out;
+  std::copy(in, in + _offset, std::back_inserter(out));
+  for (int i = 0; i < _num_units; ++i) {
+    //    decode_golay24(&encoded);
+  }
+  std::copy(in + _offset + 24 * _num_units, in + in_size,
```
reiinakano: I think it might be easier to read to do something like:
```
out = copy of in
for (i=offset; i<size-24; i+=24) {
  error = decode(out[i])
}
```

gr-starcoder/lib/golay24.c
```diff
@@ -0,0 +1,119 @@
...
+
+#include <stdint.h>
+
+#include <volk/volk.h>
+
+#include "golay24.h"
+
+#define N 12
+
+static const uint32_t H[N] = { 0x8008ed, 0x4001db, 0x2003b5, 0x100769,
```
reiinakano: One of the things I wanted to do is make this a configurable parameter. It could take an array or maybe we have an enum of the most common settings for this parameter. This constant is the `H` in https://www.maplesoft.com/applications/view.aspx?sid=1757&view=html, but I think we can generate it from a 12 bit input, which in the link is `b1` (so our configurable parameter will be `b1`. I am not sure if there is an official term for this).

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,96 @@
...
+      if (*p != 0) {
+        word |= 1;
+      }
+      p++;
+    }
+    decode_golay24(&word);
+
+    uint32_t mask = 1 << 11;
+    for (int j = 0; j < 12; j++) {
+      out.push_back(word & mask ? 1 : 0);
```
kenichi-fukushima: Is this the correct representation of unpacked binary stream?

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,96 @@
...
+      mask = mask >> 1;
+    }
+  }
+
+  std::copy(in + offset_ + 24 * num_units_, in + in_size,
+            std::back_inserter(out));
+
+  message_port_pub(
+      pmt::mp("out"),
+      pmt::cons(pmt::PMT_NIL, pmt::init_u8vector(out.size(), out)));
```
kenichi-fukushima: What should the key of output dictionary be? Can I just forward input key?

gr-starcoder/lib/golay24.c
anuraaga: For borrowed code, we put both the original and our license (or at least should be). Any edits we make to the fork would be under our license (we can't write code but license it to someone totally unrelated to us ;) ).

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -0,0 +1,98 @@
...
+    for (int j = 0; j < 24; j++) {
+      if (j != 0) {
+        word = word << 1;
+      }
+      if (*p != 0) {
+        word |= 1;
+      }
+      p++;
+    }
+    if (decode_golay24(&word) < 0) {
```
reiinakano: Let's do `GR_LOG_DEBUG` whenever golay decoding succeeds (similar to what the Reed Solomon decoder block does), logging the number of bits corrected.

gr-starcoder/lib/golay_decoder_impl.cc
reiinakano: Yes, it seems correct. Is `push_back` similar to `append` in python? 

Anyway, I think we should still include the parity bits in the final output to mirror what RS decoder does.

gr-starcoder/lib/golay_decoder_impl.cc
kenichi-fukushima: I'm a bit concerned that it might generate too much log data. Do you have an assessment about rate of golay encoded data that we receive?

gr-starcoder/lib/golay_decoder_impl.cc
kenichi-fukushima: Yes. `push_back` is `append` in other languages.

I have no idea what you mean by parity bits. Can you refer to code/document?

reiinakano: Oh, parity bits are the extra error-correcting bits in the error correcting code. I believe in Golay this is <12 parity bits><12 data bits>

gr-starcoder/lib/golay_decoder_impl.cc
reiinakano: Hmm, okay I guess it's fine to leave this as is.

gr-starcoder/lib/golay_decoder_impl.cc
reiinakano: Actually, I'm not sure anymore if we should or should not strip it out. I think it will be easy to modify anyway if we find that we need to keep the parity bits.

gr-starcoder/lib/golay_decoder_impl.cc
reiinakano: Good point. I'm not sure if there's a guarantee actually. Our [command_source_impl block](https://github.com/infostellarinc/starcoder/blob/master/gr-starcoder/lib/command_source_impl.h) works fine without it though. Anyway, not important so we can leave it as is.

kenichi-fukushima:
Added logging assuming date rate of satellite communication is not very high and a message usually contains at most one or two Golay block.

The default log level of GnuRadio is debug and it seems like we don't override the value.
https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/gnuradio-runtime.conf.in

gr-starcoder/lib/golay_decoder_impl.cc
```diff
@@ -75,9 +75,13 @@ void golay_decoder_impl::msg_handler(pmt::pmt_t pmt_msg) {
       }
       p++;
     }
-    if (decode_golay24(&word) < 0) {
+    int num_corrected_bits = decode_golay24(&word);
+    if (num_corrected_bits < 0) {
+      GR_LOG_INFO(d_logger, "Failed to decode a Golay encoded message.");
```
reiinakano: This should also be debug since we can expect we'll be decoding a lot of fake packets via false positive sync word matching.

kenichi-fukushima: Ah, good point!

PR #113, opened 2018-12-04
FSK AX100 flowgraph

reiinakano:
fsk_ax100 grc

Needs the gr-sattools OOT to be in the GNURadio build to work, but here's the flowgraph.


vaelen:
The low pass filter should probably change too depending on the bit rate.

PR #114, opened 2018-12-04
Add gr-sattools in build

reiinakano:


gr-recipes/gr-sattools.lwr
```diff
@@ -0,0 +1,7 @@
+category: common
+depends:
+- gnuradio
+description: GNU Radio decoder for AX100
+gitbranch: master
+inherit: cmake
+source: git+https://github.com/infostellarinc/gr-sattools
```
anuraaga: Since the PR was merged can we use the upstream one?

reiinakano: We made a small change in our fork https://github.com/infostellarinc/gr-sattools/pull/3 that I'm not sure will be merged upstream. Guess I should open a PR but we can change it to upstream easily if it's merged anyway.

PR #117, opened 2018-12-08
typo

reiinakano:
use a correct ax25 header now


reiinakano:
merging but @vaelen double check the header if you want

PR #118, opened 2018-12-08
Fix preamble lebgth

reiinakano:
Our previous flowgraph set the preamble and postamble length at 16. The flowgraph they sent us has it set to 192 and 64. The calculation we have in SR for timing the turning on and off the switches does not reflect this. I am not sure if our margin of allowance in timing is enough. I think we want to push this in before they try transmitting.

### DIFF
--- a/flowgraphs/fsk_9600_ax25_transceiver.grc
+++ b/flowgraphs/fsk_9600_ax25_transceiver.grc
@@ -2019,11 +2019,11 @@
     </param>
     <param>
       <key>postamble_bytes</key>
-      <value>64</value>
+      <value>16</value>
     </param>
     <param>
       <key>preamble_bytes</key>
-      <value>192</value>
+      <value>16</value>
     </param>
   </block>
   <block>


reiinakano:
Sorry for the late PR, I just thought of this.

reiinakano:
Well we calculate it based on max packet size of 256 bytes and theyre not transmitting that much so I guess it's fine

anuraaga:
Does the length have any other implications than timing?

reiinakano:
Right now their usual packet size is 56 bytes including header. Add in 2 bytes for crc and 256 bytes for pre+postamble, that is 314 bytes.

For 9600bps that is 0.262 seconds. Our current margin is 0.26 seconds for ax.25 packets..

anuraaga:
What do you think about using middle ground values to fall in our margin while giving the receiver more time to sync as necessary?

reiinakano:
Oh, that's perfect lol. Changing now

reiinakano:
Closing because of the change in SR

PR #119, opened 2018-12-10
Copy only needed files to starcoder image.

anuraaga:
I verified that starcoder starts up but it's hard to actually be confident without also checking a flowgraph.

The raw image goes from 3.5G to 2.9G. I didn't check compressed size, but I guess the difference is much less since much of what's left out is files like `docs`.

### DIFF
--- a/tools/builder/Dockerfile
+++ b/tools/builder/Dockerfile
@@ -15,7 +15,20 @@ RUN cd /workspace/starcoder && ./gradlew install
 
 FROM debian:9-slim
 
-COPY --from=0 /root/.gradle/curiostack/gnuradio /root/.gradle/curiostack/gnuradio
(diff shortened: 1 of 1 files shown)


tools/builder/Dockerfile
```diff
@@ -15,7 +15,20 @@ RUN cd /workspace/starcoder && ./gradlew install
 
 FROM debian:9-slim
 
-COPY --from=0 /root/.gradle/curiostack/gnuradio /root/.gradle/curiostack/gnuradio
+COPY --from=0 /root/.gradle/curiostack/gnuradio/4.4.10/etc /root/.gradle/curiostack/gnuradio/4.4.10/etc
```
reiinakano: Where is the 4.4.10 from? I'm wondering if this is something we'll need to change when we update some package or something.

anuraaga: It's the version of conda - will see if there's a simple way to clean up the version specification in another PR

PR #121, opened 2018-12-20
Script for receiver evaluation

reiinakano:
Helper script for running receiver evaluation with a configuration file. View README for details on how it is supposed to be used.

### DIFF
--- a/gr-starcoder_utils/python/qa_prbs_sink_pdu.py
+++ b/gr-starcoder_utils/python/qa_prbs_sink_pdu.py
@@ -280,11 +280,12 @@ class qa_prbs_sink_pdu (gr_unittest.TestCase):
         for packet in self.packets:
             snk.to_basic_block()._post(pmt.intern('all'), pmt.cons(pmt.PMT_NIL, pmt.to_pmt(packet)))
         for i, packet in enumerate(self.packets):
-            if i == 10:
+            if i == 0:
                 packet = np.copy(packet)
                 # This makes the packet bigger than NUM_GENERATED_PACKETS
-                packet[0] = 0xff
-                packet[1] = 0xff
+                packet[0] = 0x90
+                packet[1] = 0x1
+                print(packet)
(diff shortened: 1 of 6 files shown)


gr-starcoder_utils/python/qa_prbs_sink_pdu.py
```diff
@@ -266,6 +266,44 @@ def test_006_some_dupes(self):
         expected[1] += 1
         self.assertTrue(np.array_equal(snk.collected_packets, expected))
 
+    def test_007_some_packets_with_wrong_headers(self):
+        snk = prbs_sink_pdu(self.PACKET_LENGTH_BYTES*8, self.PRBS_RESET_LEN, self.NUM_GENERATED_PACKETS)
+
+        # We just need something connected to the trimmer block for
+        # the flowgraph to compile, but we'll post messages to it directly
+        src = blocks.message_strobe(pmt.PMT_NIL, 9999999)
```
kenichi-fukushima: not needed?

gr-starcoder_utils/python/qa_prbs_sink_pdu.py
```diff
@@ -266,6 +266,44 @@ def test_006_some_dupes(self):
...
+        self.tb.start()
+        for packet in self.packets:
+            snk.to_basic_block()._post(pmt.intern('all'), pmt.cons(pmt.PMT_NIL, pmt.to_pmt(packet)))
+        for i, packet in enumerate(self.packets):
+            if i == 0:
+                packet = np.copy(packet)
+                # This makes the packet bigger than NUM_GENERATED_PACKETS
+                packet[0] = 0x90
+                packet[1] = 0x1
+                print(packet)
```
kenichi-fukushima: remove this debug print or use the logger

gr-starcoder_utils/python/qa_prbs_sink_pdu.py
reiinakano: Just remembered why this is needed. The block needs something connected to it. The golay decoder and other blocks don't need the dummy block since they are connected to a sink. In this case, the block to be tested *is* a sink

PR #122, opened 2019-01-07
add preamble of alternating 1s and 0s between packets

reiinakano:
generatorg grc


reiinakano:
Usually preambles are for giving the receiver time to synchronize with the bit rate before receiving a packet.

reiinakano:
Closing for now since setting preamble_bits to 0 breaks

PR #123, opened 2019-01-07
Allow enabling and disabling of blocks

reiinakano:
evaluatorr grc

This PR introduces hierarchical blocks that wrap the Viterbi decoder, Additive Scrambler, and RS decoder blocks so we can enable and disable them at runtime.

### DIFF
--- a/gr-starcoder_utils/python/viterbi_decoder.py
+++ b/gr-starcoder_utils/python/viterbi_decoder.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2019 Infostellar.
+# 
+# This is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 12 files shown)


gr-starcoder_utils/python/viterbi_decoder.py
```diff
@@ -0,0 +1,73 @@
...
+class viterbi_decoder(gr.hier_block2):
+    """
+    This hierarchical block simply wraps around the FEC Async Decoder block in core
+    GNU Radio. It adds an additional parameter `enabled` to determine whether to
+    decode the packet or let it through without change.
+    """
+    def __init__(self, cc_decoder_definition, packed, rev_packing, mtu_bytes, enabled):
+        """
+
+        :param cc_decoder_definition: CC decoder definition created by fec.cc_decoder.make
```
kenichi-fukushima: what's CC?

reiinakano: convolutional coding I think. I just copied the name from the original block

PR #124, opened 2019-01-07
Allow enabling and disabling of blocks for generator flowgraph

reiinakano:
generatorr grc

Same thing as the previous PR. Wrapped some blocks in hierarchical blocks so we can enable and disable at runtime.

### DIFF
--- a/gr-starcoder_utils/python/conv_encoder.py
+++ b/gr-starcoder_utils/python/conv_encoder.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# 
+# Copyright 2019 Infostellar.
+# 
+# This is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3, or (at your option)
+# any later version.
+# 
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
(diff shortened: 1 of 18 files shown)


gr-starcoder_utils/python/rs_encoder.py
```diff
@@ -0,0 +1,50 @@
...
+    gr-satellites. It adds an additional parameter `enabled` to determine
+    whether to decode the packet or let it through without change.
+    """
+    def __init__(self, basis, enabled):
+        gr.hier_block2.__init__(self,
+            "rs_encoder",
+            gr.io_signature(0, 0, 0),  # Input signature
+            gr.io_signature(0, 0, 0))  # Output signature
+
+        self.log = gr.logger("log")
```
kenichi-fukushima: not used

gr-starcoder_utils/python/conv_encoder.py
```diff
@@ -0,0 +1,48 @@
...
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+from gnuradio import gr
+from gnuradio import fec
+from pmt_pass import pmt_pass
+
+
+class conv_encoder(gr.hier_block2):
```
kenichi-fukushima: Don't abbreviate unless this is a well-established convention in GnuRadio or DSP in general. I can't tell whether `conv` means conversion or convolution.

Also why isn't it named after the wrapped block?

reiinakano:
> I started to suspect you might use some meta-programming technique to generate these simple wrappers.

Let me think about this. It's not very hard to do in Python, but I'm not sure how to make sure we can extend this to GNURadio Companion. Let me know if you have any suggestions.

reiinakano:
Closing, I think I thought of a way to make this generic.

PR #126, opened 2019-01-08
PMT selector

reiinakano:
reverting the commit that reverts the accidental commit = the original commit...

### Generator 
generatorrr grc

### Evaluator
evaluator rrgrc

This introduces PMT Selector, which simply routes received messages across a certain path.


gr-starcoder_utils/python/pmt_selector.py
```diff
@@ -0,0 +1,50 @@
...
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+# 
+
+from gnuradio import gr
+import pmt
+
+
+class pmt_selector(gr.sync_block):
```
kenichi-fukushima: Seems `pdu` is the appropriate term here. Other blocks use `pdu` too like PDU to tagged stream.
https://wiki.gnuradio.org/index.php/Guided_Tutorial_Programming_Topics#5.3.1_PDUs

Also in computing in general `select` specifically means selecting from multiple input channels like selecting from multiple channels in a for loop in Go. This block is selecting from multiple output channels so another term is preferred to avoid confusion. How about simply `router`?

reiinakano: PDU is a special case of PMT. Blocks with PDU in their name actually require the inputs to be PDUs. However, this block can handle any generic PMT.

reiinakano: Actually, message is probably the correct term.

PR #128, opened 2019-01-08
Add modulators to Generator

reiinakano:
generatorrrrr grc

Allows us to switch between FSK, BPSK, and QPSK.


gr-starcoder_utils/apps/receiver_evaluation/mod_mapping.py
```diff
@@ -0,0 +1,26 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+
+def mod_string_to_id(x):
+  mapping = {
+    'FSK': 0,
```
kenichi-fukushima: This mapping is only valid for a specific configuration of flowgraph. We should avoid hardcoding it if possible. 

How about this approach?
- Make it so `message_router`'s constructor takes a list of output port names as a parameter. Actual value of this parameter is determined when a flowgraph is written.
- `message_router` maintains a mapping from names to ports.
- The port name to be selected among choices is another constructor parameter but is not determined until a flowgraph is run.

reiinakano: Hmm, I'm not using message_router for switching between modulations. I'm using Selector block built in to GNU Radio

kenichi-fukushima: Oh, sorry. Never mind.

kenichi-fukushima: Hardcoding the mapping should be fine provided the code is within `receiver_evaluation` directory.

I didn't know GNU Radio provides `Selector` block for stream. If I had known it use of term `selector` in your previous PR would have sounded reasonable to me.

reiinakano:
The only downside of the GNURadio `Selector` block is it's deprecated :(. As far as I can tell, the only reason for this is because switching ports *during* runtime can have undefined effects. But in our case, we set the ports at the beginning and never change them throughout, so this is probably fine.

PR #130, opened 2019-01-11
File generator

reiinakano:
file_generator grc

An optional flowgraph for running an IQ file directly instead of generating it on the fly. This is for running the tool on low power devices that cannot handle modulation and transmission at the same time

### DIFF
--- a/gr-starcoder_utils/apps/.gitignore
+++ b/gr-starcoder_utils/apps/.gitignore
@@ -1,2 +1,3 @@
 receiver_evaluation/evaluator.py
 receiver_evaluation/generator.py
+receiver_evaluation/file_generator.py


gr-starcoder_utils/apps/receiver_evaluation/mod_mapping.py
```diff
@@ -1,22 +1,4 @@
...
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this software; see the file COPYING.  If not, write to
-# the Free Software Foundation, Inc., 51 Franklin Street,
-# Boston, MA 02110-1301, USA.
-#
-
-
+# this module will be imported in the into your flowgraph
```
kenichi-fukushima: Unintentional?

reiinakano:
I realized that gnuradio does this automatically when compiling a module
into a flowgraph. Probably better to commit it so the diff will be cleaner.

reiinakano:
No, the script is indeed written by me, and not in the .grc file. I only
point to the file in the .grc. gnuradio for some inexplicable reason then
formats it that way everytime a flowgraph is compiled.

PR #136, opened 2019-03-11
fsk demodulator fix init

CanoRoger:
FSK demodulator upgraded with newer Symbol Sync module and improved Low pass filter parameters.

Configured parameters tested for optimum GFSK 9600bps performance. Capable (but still not implemented) of different bitrate as well. Tests with AISTECH passes shows +40% performance in successful received packets.

Other changes: AX25 transmitted postamble = 4 on fsk_ax25_9600_transceiver.

There are 3 modified flowgraphs:
-fsk_9600:
fsk_9600
-fsk_9600_ax25_transceiver:
fsk_9600_ax25_transceiver
-fsk_ax100:
fsk_ax100

### DIFF
--- a/flowgraphs/fsk_9600.grc
+++ b/flowgraphs/fsk_9600.grc
@@ -10,7 +10,7 @@
     </param>
     <param>
       <key>window_size</key>
-      <value></value>
+      <value>(1500, 1250)</value>
     </param>
     <param>
       <key>category</key>
@@ -101,7 +101,7 @@
     </param>
     <param>
       <key>_coordinate</key>
-      <value>(8, 396)</value>
+      <value>(8, 372)</value>
     </param>
     <param>
       <key>_rotation</key>
@@ -128,7 +128,34 @@
     </param>
     <param>
       <key>_coordinate</key>
-      <value>(8, 332)</value>
+      <value>(8, 508)</value>
+    </param>
+    <param>
+      <key>_rotation</key>
+      <value>180</value>
+    </param>
+    <param>
+      <key>id</key>
+      <value>nfilts</value>
+    </param>
+    <param>
+      <key>value</key>
+      <value>16</value>
+    </param>
+  </block>
+  <block>
+    <key>variable</key>
+    <param>
+      <key>comment</key>
+      <value></value>
+    </param>
+    <param>
+      <key>_enabled</key>
+      <value>1</value>
+    </param>
+    <param>
+      <key>_coordinate</key>
+      <value>(80, 508)</value>
     </param>
     <param>
       <key>_rotation</key>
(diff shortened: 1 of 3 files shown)


reiinakano:
I have 2 questions about this:

1. You say that variable bit rates are not yet supported. This is probably true for fsk_9600 and fsk_9600_ax25_transceiver but those flowgraphs do have a variable called "baud_rate". Are you using this variable instead of hardcoding 9600bps? In the case of fsk_ax100, variable bit rates are already supported since "baud_rate" is a parameter. Are you using this parameter?

2. Is it correct that the set of blocks after DC Blocker form an RMS AGC? Have you experimented with the other AGC blocks built into GNU Radio? The reason I'm asking is because I've had trouble with a bug in the Divide block and non-deterministic results (see http://lists.gnu.org/archive/html/discuss-gnuradio/2018-10/msg00005.html) when I was trying to decode Spire data. Switching to one of the built-in GNURadio AGCs led to better (and deterministic) results for me, but it might not be the case for this flowgraph.

flowgraphs/fsk_9600.grc
```diff
@@ -155,19 +182,19 @@
...
     </param>
     <param>
       <key>id</key>
-      <value>lpf_cutoff</value>
+      <value>source_filepath</value>
     </param>
     <param>
       <key>value</key>
-      <value>8000</value>
+      <value>"/home/rei/sampleAR2300IQ/full.bin"</value>
```
koncha99: remove a value from personal environment.

flowgraphs/fsk_ax100.grc
```diff
@@ -143,6 +170,60 @@
       <value>"/home/rei/sampleAR2300IQ/full.bin"</value>
```
koncha99: same as above

CanoRoger:
1. Yes, I am using baud_rate. I didn't realize it in the fsk_ax100. So yes, it is compatible with variable baud_rate.

2. It is the contents of the gr-satellite's hier. RMS_AGC but converted to handle float streams. I was not aware of this issue. I remember into a early stage trying to use the other AGCs but this one seemed to perform the best. Let me test it again with this final configuration.

reiinakano:
2. It's fine if that one performs the best. Was just wondering if you tried the other AGCs

flowgraphs/fsk_9600.grc
reiinakano: Hmm, we currently have this variable in all our flowgraphs. They're part of a disabled block that I like to keep around just for debugging. Do you think we should remove them? In any case, I think we can deal with them in a separate PR.

kenichi-fukushima: It's totally fine to have such a developer feature in the shared repository. But ideally actual configuration of the feature is in separate files just like .bashrc and IntelliJ workspace. Then you can use `.gitignore`. Is there a way to provide parameters from `~/.gnuradio/config.conf` or a separate flow graph?

reiinakano:
Let's replace those hardcoded variables with https://www.ece.uvic.ca/~ece350/grc_doc/ar01s04s02.html in a separate PR

PR #137, opened 2019-04-09
Graceful shutdown with SIGHUP

koncha99:
Currently Starcoder will exit immediately if it get signals.
This change enables that Starcoder wait for finishing flow graph execution before exiting if it get SIGHUP.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -167,9 +167,9 @@ var serveCmd = &cobra.Command{
 	},
 }
 
-func doRestart(starcoder *server.StarCoder, s *grpc.Server) {
+func doRestart(starcoder *server.Starcoder, s *grpc.Server) {
 	for {
-		if starcoder.getCurrentFlowgraphCount() == 0 {
+		if starcoder.GetCurrentFlowgraphCount() == 0 {
 			starcoder.Close()
 			s.GracefulStop()
 			return
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -140,6 +140,9 @@ var serveCmd = &cobra.Command{
 				select {
 				case sig := <-sigs:
 					switch sig {
+					case syscall.SIGHUP:
+						log.Infof("Starcoder will restart.")
+						go doRestart(starcoder, s)
 					case syscall.SIGINT:
 						fallthrough
 					case syscall.SIGTERM:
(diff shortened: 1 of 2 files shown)


server/starcoder.go
```diff
@@ -42,19 +42,21 @@ import (
...
-	log                       *zap.SugaredLogger
-	silencedCommandBlocks     map[string]bool
+	flowgraphDir                  string
+	gilState                      python.PyGILState
+	threadState                   *python.PyThreadState
+	streamHandlers                map[*streamHandler]bool // registered stream handlers
+	registerStreamHandler         chan *streamHandler
+	deregisterStreamHandler       chan *streamHandler
+	closeAllStreamsChannel        chan chan bool
+	requestFlowgraphCountChannel  chan struct{}
```
reiinakano: I think the usual pattern is to pass a `chan chan int` so you only need to define one channel in the struct here. Wdyt @vaelen ?

kenichi-fukushima: Yeah, you are right. I overlooked there were requestChannel and responseChannel.

cmd/serve.go
```diff
@@ -164,6 +167,17 @@ var serveCmd = &cobra.Command{
 	},
 }
 
+func doRestart(starcoder *server.Starcoder, s *grpc.Server) {
+	for {
```
hoshir: Since I don't usually see PR of starcoder, I don't full context of what the code is doing. So please receive my just as a reference.

I don't think we should poll manually. What we want to do here is to wait until all flowgraph is detached. For that purpose, I come up with two idioms:
1. Use WaitGroup
WaitGroup perfectly matches the purpose which waits until all resources are freed by Done method. starcoder can return WaitGroup for that purpose.
2. Use "done" channel
GetCurrentFlowgraphCount returns channel which is closed when all resources are detached. The go routine just wait for the close with for-select loop. The benefit of using "done" channel is you can close the channel and stop the go routine by yourself in the server.go, such case like Ctl+C is pressed.

vaelen: Yes, @hoshir is right.  WaitGroup does exactly what you want here.  Before calling `go f()` you call `wg.Add(1)` to add to the wait group, then in this code you just call `wg.Wait()` and it will return when the count hits 0.

PR #140, opened 2019-04-24
STELLAR-2523 Add tcp source

reiinakano:


reiinakano:
Since CI always times out, I'll merge once approved.

gr-starcoder/python/radio_source.py
```diff
@@ -59,6 +62,13 @@ def __init__(self, radio, device_address, samp_rate, center_freq, gain, antenna)
             self.radio_block.set_center_freq(center_freq, 0)
             self.radio_block.set_gain(gain, 0)
             self.radio_block.set_antenna(antenna, 0)
+        elif radio == "TCP":
+            self.radio_block = grc_blks2.tcp_source(
+                itemsize=gr.sizeof_gr_complex*1,
+                addr=device_address,
+                port=port_number,
```
koncha99: If address or port number is not assigned, does this raise any warn or error ?

reiinakano: It will just fail if the address or port cannot be connected to.

PR #141, opened 2019-04-25
Exit Starcoder after TCP stream to work around TCP source block bug.

reiinakano:
cc @vaelen

### DIFF
--- a/server/starcoder.go
+++ b/server/starcoder.go
@@ -56,6 +56,9 @@ type Starcoder struct {
 	compileLock               sync.Mutex
 	log                       *zap.SugaredLogger
 	silencedCommandBlocks     map[string]bool
+
+	// If this is true, Starcoder will restart after a stream is finished. This is a hack.
+	killAfterStream bool
 }
 
 type moduleAndClassNames struct {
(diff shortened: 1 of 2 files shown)


server/starcoder.go
```diff
@@ -56,6 +56,9 @@ type Starcoder struct {
 	compileLock               sync.Mutex
 	log                       *zap.SugaredLogger
 	silencedCommandBlocks     map[string]bool
+
+	// If this is true, Starcoder will restart after a stream is finished. This is a hack.
```
vaelen:
```suggestion
	// If this is true, Starcoder will restart after a stream is finished. This is a workaround for a bug in the TCP source block.
```

server/starcoder.go
```diff
@@ -683,6 +690,12 @@ func fillDictWithParameters(dict *python.PyObject, params []*pb.StartFlowgraphRe
 			switch v := param.GetValue().GetVal().(type) {
 			case *pb.Value_StringValue:
 				convertedValue = python.PyString_FromString(v.StringValue)
+
+				// If radio is TCP, restart Starcoder after this stream. This is a hack.
```
vaelen:
```suggestion
				// If radio is TCP, restart Starcoder after this stream to work around a bug in the TCP source block.
```

PR #142, opened 2019-05-08
Correct point to restart starcoder

reiinakano:


server/starcoder.go
```diff
@@ -427,6 +423,14 @@ func (s *Starcoder) RunFlowgraph(stream pb.Starcoder_RunFlowgraphServer) error {
 		for _, q := range flowgraphProps.observableCQueues {
 			q.Delete()
 		}
+
+		if s.killAfterStream {
+			go func() {
+				// Give it some time to properly close the stream.
+				time.Sleep(time.Second * 10)
```
vaelen: I suggest putting this value in a constant.

PR #143, opened 2019-05-09
FSK CCSDS flowgraph

CanoRoger:
Created flowgraph for reception of GFSK CCSDS UHF downlink.

Viterbi standard r=1/2 k=7 with bit inversion.
ASM syncword employed.
CCSDS descrambling.
CCSDS Reed Solomon in Dual basis and interleaving = 4.

The "Enqueue Message Sink" is named = fsk_ccsds_decoded

Screenshot from 2019-05-09 15-06-10


reiinakano:
* Can you change the two hierarchical blocks you're using: `Sync and Create PDU` and `CCSDS descrambler` to the actual contents of those blocks? We currently don't compile hierarchical blocks so we can't use them.

* How is 11.7188 calculated? This seems to be a hardcoded value based on 9600 baud and 112500 sampling rate? The sampling rate will be variable, in fact, for one of our groundstations, it is 120k

reiinakano:
Nvm about 11.7188, I see it's indeed dependent on target_samp_rate

CanoRoger:
Done changing the hierarchicals. The 11.7188 is computed using a curve approximation (symmetric sigmoidal) over several optimum values for different baud_rate vs. sampling_rate that I found empirically.

The parameters, apart from the used normally are the following:
-baud_rate (9600)
-RS interleaving (4)
-syncword ("00011010110011111111110000011101")
-syncword_threshold (4)

The viterbi decoder seems to only support k=7 and 1/2. Polynomials can be inverted or negative (which would define the different versions of viterbi people uses). I guess that "NOT" at the output may have something to do with this fact.

The RS dual basis/conventional is a drop-down list on the block settings. Not sure how this can be parameterized.

vaelen:
Don't forget to update the screenshot in the PR description.

PR #144, opened 2019-05-10
Compile gr-satellites hierarchical blocks 

reiinakano:
Compile gr-satellites hierarchical blocks so we can use them as-is in flowgraphs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -380,8 +380,16 @@ task installStarcoder(type: Copy) {
     into "${gnuradioRoot}/bin"
 }
 
+task compileGrSatellitesHierarchicalBlocks(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -380,8 +380,16 @@ task installStarcoder(type: Copy) {
     into "${gnuradioRoot}/bin"
 }
 
+task compileGrSatellitesHierarchicalBlocks(type: org.curioswitch.gradle.conda.tasks.CondaExecTask) {
+    dependsOn setupPrefix
+
+    condaName 'gnuradio'
+
+    command "git clone https://github.com/daniestevez/gr-satellites.git /tmp/gr-satellites && cd /tmp/gr-satellites/ && ./compile_hierarchical.sh"
```
reiinakano: Unsure if dumping this into /tmp is the best way

PR #147, opened 2019-05-23
STELLAR-2628 Strip AX.25 headers

reiinakano:


vaelen:
Please make sure there is a Jira ticket so we can track what is in the release properly.

PR #154, opened 2019-06-03
BPSK AX.25

CanoRoger:
Created flowgraph for low datarate BPSK demodulation and AX.25 decoding using core and starcoder blocks only.

Tested on 9600bps and 1200bps with satellites FengMaNiu-1, Zhou Enlai and NJUST-1.

Boolean parameters accepted for G3RUH scrambling and AX.25 header stripping.

bpsk ax25


reiinakano:
You're missing `radio_port_number`. I think you should update your Starcoder repository to get the new parameters

PR #156, opened 2019-06-12
STELLAR-2707 Starpass API block

reiinakano:


kenichi-fukushima:
Let's avoid mention of "StarPass" throughout the code. Instead we should choose "Ground Station API", "the server" or similar words depending on the context. Starcoder is not a client of StarPass but of the Ground Station API. The API does not assume a component that serves it.

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+import grpc
+from google.auth import jwt as google_auth_jwt
+from google.auth.transport import grpc as google_auth_transport_grpc
+from stellarstation.api.v1.groundstation import groundstation_pb2
+from stellarstation.api.v1.groundstation import groundstation_pb2_grpc
+from stellarstation.api.v1 import transport_pb2
+
+class groundstation_api(gr.sync_block):
+    """
+    This block communicates with the Ground Station API and performs two functions: retrieving commands from the
```
kenichi-fukushima: `communicates with the Ground Station API server`

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+    groundstation and sending telemetry to the groundstation.
+    Input PMTs are expected to be blobs containing the serialized string corresponding to a transport.Telemetry
+    protocol buffer.
```
kenichi-fukushima: `message.`

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+    https://github.com/infostellarinc/stellarstation-api/blob/0.3.0/api/src/main/proto/stellarstation/api/v1/transport.proto#L63
+    Commands received from the groundstation are sent to the rest of the flowgraph as uint8 PDUs.
```
kenichi-fukushima: This is a very important design decision that this block communicates only telemetry and commands with other blocks. How do you compare it with an alternative of it sends GroundStationStreamResponse and receives GroundStationStreamRequest?

reiinakano: I want this block to handle as much of the Ground Station API logic as possible. 

For passing on commands:
* Other GNU Radio blocks work out of the box with uint8 PDUs. This logic of parsing out the command bytes from GroundStationStreamResponse needs to live somewhere (here or upstream blocks). Currently, I see no reason for it to be outside this block.

For receiving telemetry:
* Here, we're expecting a `transport.Telemetry` since we have to tag the correct `Framing` outside this block. I will probably be introducing a new block that converts u8vector PDUs (usual output of GNURadio blocks) to the appropriate `transport.Telemetry` data type. Alternatively, there could be multiple inputs to this block, one for each `Framing`.

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+        while not self.get_stopped():
+            try:
+                telemetry = self.queue.get(timeout=1)
+                yield self._telemetry_to_stream_request(telemetry)
+                self.queue.task_done()
+            except Empty:
+                continue
+
+        while not self.queue.empty():
+            telemetry = self.queue.get()
```
koncha99: What could happen if timeout isn't specified ?

reiinakano: We check if the queue is empty before reading from it, and since this is the only thread that reads from the queue, it should be fine.

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+# 
+# You should have received a copy of the GNU General Public License
+# along with this software; see the file COPYING.  If not, write to
+# the Free Software Foundation, Inc., 51 Franklin Street,
+# Boston, MA 02110-1301, USA.
+#
+
+from Queue import Queue, Empty
```
kenichi-fukushima: Just curious. Which queue is this?

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+            self.queue.task_done()
+
+    # This function runs asynchronously and starts the bidirectional stream.
+    # All commands/messages from the groundstation are sent to the appropriate PMT ports.
+    def handle_stream(self):
+        for response in self.api_client.OpenGroundStationStream(self.generate_request()):
+            if response.HasField("satellite_commands"):
+                for command in response.satellite_commands.command:
+                    if self.verbose:
+                        print("Sending command", command[:20])
```
kenichi-fukushima: `gr_log`?

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,174 @@
...
+                self.queue.task_done()
+            except Empty:
+                continue
+
+        while not self.queue.empty():
+            telemetry = self.queue.get()
+            yield self._telemetry_to_stream_request(telemetry)
+            self.queue.task_done()
+
+    # This function runs asynchronously and starts the bidirectional stream.
```
kenichi-fukushima: `asynchronously` -> `on another thread`

gr-starcoder/python/groundstation_api.py
reiinakano: Multiple blocks means we connect to the API N times for a flowgraph. This is probably unnecessary. Multiple inputs is fine, but we'll have to change this block every time we add a new `Framing`. Wdyt?

So far I've been ignoring the other fields (downlink_frequency_hz, time_first_byte_received, time_last_byte_received, and frame_header). frame_header is particularly annoying. There's no way to fill this in other than via a custom block that outputs transport.Telemetry. The rest can be filled in automatically by this API block.

gr-starcoder/python/groundstation_api.py
reiinakano: https://docs.python.org/2.7/library/queue.html

gr-starcoder/python/groundstation_api.py
kenichi-fukushima: I'm inclined to the API block receiving telemetry bytes provided most fields will be set by the API block. We want to confuse external developers by exposing details that they don't have to know.

W.r.t., frame selection, the most important point is adding a new framing is backward-compatible. And here I came to a new question. When an external developer gives a flowgraph to us, how is the API block represented? If they use a generic sink/source block and we replace it with the API block, how do they specify a framing?

reiinakano: > I'm inclined to the API block receiving telemetry bytes provided most fields will be set by the API block. We want to confuse external developers by exposing details that they don't have to know.

How should we handle `frame_header`? Add another input that takes in a full serialized `transport.Telemetry` in special cases?

> generic sink/source block 

Isn't this out of scope for this block? This block passes commands and telemetry to/from Starpass. The generic sink/source blocks we talked about are for passing IQ data to/from radios directly within the flowgraph.

gr-starcoder/python/groundstation_api.py
```diff
@@ -0,0 +1,195 @@
...
+from google.auth.transport import grpc as google_auth_transport_grpc
+from stellarstation.api.v1.groundstation import groundstation_pb2
+from stellarstation.api.v1.groundstation import groundstation_pb2_grpc
+from stellarstation.api.v1 import transport_pb2
+
+class groundstation_api(gr.sync_block):
+    """
+    This block communicates with the Ground Station API server and performs two functions: retrieving commands from the
+    groundstation and sending telemetry to the groundstation.
+    Input PMTs are expected to be blobs containing the serialized string corresponding to a transport.Telemetry
```
kenichi-fukushima: Update the comment :)

PR #159, opened 2019-06-26
Fix bugs in groundstation_api_doppler block

reiinakano:
Found via testing with fake server


gr-starcoder/python/groundstation_api_doppler.py
```diff
@@ -113,6 +113,7 @@ def processing_loop(self):
         plan = [x for x in listPlansResponse.plan if x.plan_id == self.plan_id]
         if not plan:
             self.log.fatal("No plans matching plan ID {} found.".format(self.plan_id))
+            raise Exception('No plans matching plan ID {} found'.format(self.plan_id))
```
kenichi-fukushima: Oh, `log.fatal` does not terminate the program?

reiinakano: Surprisingly no

PR #163, opened 2019-07-08
STELLAR-2792 Flowgraph startup script

reiinakano:
This script is used to run/compile-and-run a flowgraph with the given parameters and optionally "collect" and send any files generated by the flowgraph to the ground station API (eg waterfall plots).

### DIFF
--- a/tools/builder/start_flowgraph.py
+++ b/tools/builder/start_flowgraph.py
@@ -18,6 +18,16 @@
 # the Free Software Foundation, Inc., 51 Franklin Street,
 # Boston, MA 02110-1301, USA.
 #
+"""
+Sample command to use this script
+$ python tools/builder/start_flowgraph.py \
+    -F flowgraphs/test.grc \
+    -C /tmp/flowgraph_params.yaml \
+    -X /tmp/waterfall.png WATERFALL \
+    -U localhost:50051 \
+    -G 2 \
+    -P 3
+"""
 
 import argparse
 import json
@@ -28,6 +38,7 @@ import importlib
 import tempfile
 import yaml
 _mapping_tag = yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG
+from itertools import izip
 
 from gnuradio import gr
 
@@ -43,8 +54,10 @@ parser = argparse.ArgumentParser(description="Script to compile and run a flowgr
 
 parser.add_argument("-F", "--flowgraph", help="Location of flowgraph file", required=True)
 parser.add_argument("-C", "--config", help="Location of configuration file", required=True)
-parser.add_argument('-X','--collect', nargs='*', help='List of files to collect and send to the ground station API '
-                                                      'at the end of the pass. e.g. Waterfall plots')
+parser.add_argument('-X','--collect', nargs='*', help='List of files to collect and their respective framing type to '
+                                                      'send to the ground station API at the end of the pass. The '
+                                                      'filepath and corresponding frame type must be sent alternately'
+                                                      ' e.g. waterfall.png WATERFALL weather.jpg IMAGE_JPEG')
 
 # The below arguments are only used when there are files to collect.
 parser.add_argument('-K', '--api-key', help="Location of API key for connecting to the ground station API.")
@@ -53,6 +66,12 @@ parser.add_argument('-U', '--api-url', help="URL for ground station API")
 parser.add_argument('-G', '--ground-station-id', help="Ground Station ID", default='')
 parser.add_argument('-S', '--stream-tag', help="Stream tag for this plan", default='')
 parser.add_argument('-P', '--plan-id', help="Plan ID", default='')
+
+
+def pairwise(iterable):
+    """s -> (s0, s1), (s2, s3), (s4, s5), ..."""
+    a = iter(iterable)
+    return izip(a, a)
 
 
 def compile_flowgraph(path):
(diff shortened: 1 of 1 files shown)


tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
...
+        raise Exception('files_to_collect does not have an even number of elements. Do all files have a frame type?')
+    print("Reading flowgraph configuration..")
+    with open(args.config) as f:
+        config = yaml.load(f, Loader=yaml.FullLoader)
+        if config is None:
+            config = dict()
+    print('Flowgraph configuration', json.dumps(config, indent=2))
+
+    _, ext = os.path.splitext(args.flowgraph)
+    if ext == '.py':
```
kenichi-fukushima: Shouldn't this be the other way around?

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 Infostellar.
```
kenichi-fukushima: 2019

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
...
+from google.auth.transport import grpc as google_auth_transport_grpc
+
+from stellarstation.api.v1.groundstation import groundstation_pb2
+from stellarstation.api.v1.groundstation import groundstation_pb2_grpc
+from stellarstation.api.v1 import transport_pb2
+
+parser = argparse.ArgumentParser(description="Script to compile and run a flowgraph")
+
+parser.add_argument("-F", "--flowgraph", help="Location of flowgraph file", required=True)
+parser.add_argument("-C", "--config", help="Location of configuration file", required=True)
```
kenichi-fukushima: Please be explicit that this config is actually flowgraph parameters in both naming and comment. I though it was for configuring behavior of this script.

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
...
+    sys.path.append(py_folder)
+    module_name, _ = os.path.splitext(os.path.basename(flowgraph_py_path))
+    print("Using module name {}".format(module_name))
+    module = importlib.import_module(module_name)
+    flowgraph_class = module.__dict__[module_name]
+    assert issubclass(flowgraph_class, gr.top_block)
+
+    flowgraph = flowgraph_class(**config)
+    print("Running flowgraph. CTRL + C to stop.")
+    flowgraph.run()
```
kenichi-fukushima: When does this method return?

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if args.collect:
+        print("Collecting files: {}".format(args.collect))
+        api_client = setup_api_client(args.root_cert_path, args.api_key, args.api_url)
+        for _ in api_client.OpenGroundStationStream(generate_request(
```
kenichi-fukushima: I don't understand how this line works...  Does a grpc stream API in Python consume a generator and return a generator?

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,194 @@
...
+                        )
+                    )
+                )
+            yield stream_request
+        else:
+            print("WARNING: {} is not a file".format(path))
+
+
+def string_to_framing(x):
+    mapping = {
```
kenichi-fukushima: Use something from the protobuf library instead of defining your own map.
https://developers.google.com/protocol-buffers/docs/reference/python-generated#enum

tools/builder/start_flowgraph.py
reiinakano: No, if it's a python file there's no need to compile.

tools/builder/start_flowgraph.py
reiinakano: That is correct. The returned generator here loops over server responses to client. The consumed generator sends requests to the server.

Unfortunately the groundstation API stream we're using is bidirectional so we have to use this pattern here. If we had another API to send data from the client without listening to the server responses, this will be simpler.

tools/builder/start_flowgraph.py
reiinakano: I don't think it returns anything. It blocks until the flowgraph ends or a stop signal (CTRL+C) is received.

tools/builder/start_flowgraph.py
kenichi-fukushima: Ah, you are right.

tools/builder/start_flowgraph.py
kenichi-fukushima: That's what I wanted to know. When does the flowgraph end?

reiinakano: In normal cases, it just continuously runs until a stop signal is received. In our case I imagine this signal will be sent when we stop the Docker container running this script.

tools/builder/start_flowgraph.py
```diff
@@ -0,0 +1,184 @@
...
+        else:
+            print("WARNING: {} is not a file".format(path))
+
+
+def string_to_framing(x):
+    x = x.upper()
+    if x not in transport_pb2.__dict__:
+        raise Exception('Unknown framing type {}'.format(x))
+
+    return transport_pb2.__dict__[x]
```
kenichi-fukushima: Doesn't this work?

```
return transport_pb2.Framing.Value(x)
```

https://developers.google.com/protocol-buffers/docs/reference/python-generated#enum

reiinakano: Thanks!

PR #164, opened 2019-07-09
Add a missing linker dependency

kenichi-fukushima:


kenichi-fukushima:
@reiinakano Irreproducible is the most notable property of C/C++ :)

PR #171, opened 2019-07-26
Fix flowgraphs

reiinakano:
`waterfall_image_file_path` did not exist in some flowgraphs.


vaelen:
Can you explain what was broken with the FSK_CCSDS flowgraph?  This is being actively used by SA Catapult for TT&C.

reiinakano:
Oh, this is purely for Starpass v2. It didn't have a waterfall_image_file path parameter (previously it was directly sent instead of written to file). Functionally it should stay the same.

reiinakano:
I also increased FFT size for the waterfall to make it consistent with our other flowgraphs. For some reason it was configured differently

PR #175, opened 2019-08-19
Replaced AX.25 deframer with HDLC deframer.

rockets3600:


rockets3600:
Screenshot from 2019-08-19 18-41-54

rockets3600:
Screenshot from 2019-08-19 18-51-49

flowgraphs/fsk_9600.grc
```diff
@@ -194,7 +194,7 @@
     </param>
     <param>
       <key>value</key>
-      <value>"/home/rei/sampleAR2300IQ/full.bin"</value>
+      <value>"/home/colin/dev/iq/spaceit.bin"</value>
```
koncha99: Can you make this empty ?

CanoRoger:
Looks good. Although we may upgrade as well several issues we've identified previously.

Not sure if we want to change them now.

-The first is that the "divide" block is buggy and drops packets sometimes. So instead of having that "RMS-Multiply Cons-Add Cons-Divide" which acts as an AGC, just placing a normal GNURadio "AGC block":

With (Rate 0.1)(Reference 0.7)(Gain 1.0)(Max gain 400000)

-Also, the NRZI block can be substituted by this whole combination:
Screenshot from 2019-08-19 19-06-33
It's silly, but actually this chain consumes much less computing performance than the NRZI block...

rockets3600:
Screenshot from 2019-08-20 11-43-23

PR #181, opened 2019-09-24
Disabled PMT to PDU block to fix fsk_ax25.grc

rockets3600:


rockets3600:
Screenshot from 2019-09-24 17-23-22

After bypassing the PMT to PDU block, the flowgraph produces packets again. The assumption is that the strip AX.25 header already returns a PDU, so trying to convert it to another PDU causes it to fail.

PR #182, opened 2020-01-21
Prevent an timing issue in doppler compensation block

koncha99:
Fix a timing issue which causes an invalid argument error at `time.sleep()` by passing an negative value.

### DIFF
--- a/gr-starcoder/python/groundstation_api_doppler.py
+++ b/gr-starcoder/python/groundstation_api_doppler.py
@@ -149,7 +149,10 @@ class groundstation_api_doppler(gr.sync_block):
                 time.sleep(1)
                 if self.stopped:
                     return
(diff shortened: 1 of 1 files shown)


gr-starcoder/python/groundstation_api_doppler.py
```diff
@@ -149,7 +149,10 @@ def processing_loop(self):
                 time.sleep(1)
                 if self.stopped:
                     return
-            time.sleep(coord_time - time.time())
```
danada: I guess the issue was caused by the above `while` sleeping for more than a second?

koncha99: or the case where `coord_time - time.time()` is almost zero.

PR #185, opened 2022-08-31
Fix build

dsypniewski:
Fixes:
* pinning versions of dependencies
* removed bintray and jcenter repositories
* added missing java dependencies that were hosted on bintray


dezyh (approved):
LGTM!

